PR #1, opened 2023-09-14
add fetchData and downloadData methods

klinakerdisy:


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
```
ftarrach, line 9: can you please add empty lines before and after the heading?

README.md
```diff
@@ -187,3 +187,41 @@ cadenzaClient.on('editGeometry:ok', (event) => {
...
+Fetch data from a workbook view in CSV format.
+
+```javascript
+const response = await cadenzaClient.fetchData('{embeddingTargetId}', { 
+  accept: 'text/csv'
+});
+
+const text = await response.text();
+...
+parsedTextArray.includes('foo');
```
ftarrach, line 204: As a reader I would be very confused where `parsedTextArray` is coming from and why it would include `foo`. IMO the await two lines above is enough for an easy example

README.md
```diff
@@ -187,3 +187,41 @@ cadenzaClient.on('editGeometry:ok', (event) => {
...
+button.textContent = 'Download CSV';
+
+button.onclick = (event) => {
+  event.preventDefault();
+  cadenzaClient.downloadData('{embeddingTargetId}', {
+    accept: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+  });
+};
+
+button.click();
```
ftarrach, line 224: Shouldn't line 219-221 be enough to show the feature?

src/cadenza.js
```diff
@@ -502,6 +572,42 @@ function validGeometryType(value) {
...
+  return value === CSV || value === MS_EXCEL_2007;
+}
+
+/** @param {Promise<Response>} responsePromise
+ * @param {object} [options] - Options
+ * @param {string} [options.filename] - The media type to use for the data
+ * */
+async function download(responsePromise, { filename } = {}) {
+  const res = await responsePromise;
+  assert(res.ok, 'download failed');
```
ftarrach, line 586: why do you here throw a generic error instead of a CadenzaError?

src/cadenza.js
```diff
@@ -502,6 +572,42 @@ function validGeometryType(value) {
...
+ * */
+function getFilename(res, filename) {
+  const nameFromHeader = res.headers
+    .get('Content-Disposition')
+    ?.match(/filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/)?.[1]
+    ?.replace(/['"]/g, '');
+  const format = '.' + nameFromHeader?.split('.').at(-1);
+  return filename ? filename + format : nameFromHeader;
+}
+
```
ftarrach, line 610: imo `format` shouldn't contain a leading dot. Also, you're using the term "filename" here for two different things: the basename of the file and the complete filename.

suggestion:

```
function getFileName (response, baseName) {
  const fileNameFromHeader = …
  if (baseName) {
    const format = fileNameFromHeader?.split('.').at(-1);
    return `${baseName}.${format}`
  }
  return fileNameFromHeader;
}
```

ftarrach (changes requested):

src/cadenza.js
jkissel: Should be handled in #fetch already, see my comment there.

src/cadenza.js
```diff
@@ -502,6 +572,42 @@ function validGeometryType(value) {
...
+  const res = await responsePromise;
+  assert(res.ok, 'download failed');
+  filename = getFilename(res, filename);
+
+  const a = document.createElement('a');
+  a.href = URL.createObjectURL(new Blob([await res.arrayBuffer()]));
+  a.hidden = true;
+  a.download = filename ?? '';
+
+  document.body.appendChild(a);
```
jkissel, line 594: .append()?

src/cadenza.js
```diff
@@ -275,6 +279,72 @@ export class CadenzaClient {
...
+   * @param {object} options - Options
+   * @param {string} options.accept - The media type to use for the data
+   * @param {string} [options.filename] - An optional filename
+   * @param {AbortSignal} [options.signal] - A signal to abort the download
+   * @throws For an invalid workbook view source
+   */
+  downloadData(source, { accept, filename, signal }) {
+    this.#log('CadenzaClient#downloadData', accept);
+    assert(validMediaType(accept), 'invalid media type');
+    download(this.fetchData(source, { accept, signal }), {
```
jkissel, line 313: I had hoped that the <a download> attribute overrides the filename in the Content-Disposition header, but unfortunately that's not the case.

Loading the data in-memory (which is what happens when creating an object URL) is not OK, I think, because these files might be big.

Proposal: Add another optional "filename" parameter to the endpoint. If present, that's used instead of the generated one.

src/cadenza.js
jkissel: I hope we don't need this, see my comment in downloadData().

README.md
```diff
@@ -187,3 +187,41 @@ cadenzaClient.on('editGeometry:ok', (event) => {
...
+<small>API: [CadenzaClient#downloadData](./classes/CadenzaClient.html#downloadData)</small>
+
+Download data from a workbook view in Excel format.
+
+```javascript
+const button = document.createElement('button');
+button.textContent = 'Download CSV';
+
+button.onclick = (event) => {
+  event.preventDefault();
```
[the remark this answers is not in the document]
klinakerdisy: This is something I forgot to remove after playing around with different examples. I removed it now.

sandbox.html
```diff
@@ -68,18 +68,25 @@
...
+      createGeometry ({ embeddingTargetId, geometryType, useMapSrs }) {
+        cadenzaClient.createGeometry(embeddingTargetId, geometryType, { useMapSrs: useMapSrs === 'on' });
       },
-      editGeometry ({ embeddingTargetId, geometry }) {
-        cadenzaClient.editGeometry(embeddingTargetId, JSON.parse(geometry));
+      editGeometry ({ embeddingTargetId, geometry, useMapSrs }) {
+        cadenzaClient.editGeometry(embeddingTargetId, JSON.parse(geometry), { useMapSrs: useMapSrs === 'on' });
+      },
+      fetchData ({ embeddingTargetId, mediaType }) {
+        cadenzaClient.fetchData(embeddingTargetId, { accept: mediaType });
```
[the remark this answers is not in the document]
klinakerdisy: That's what I used it for, but it might be confusing for other devs since nothing seems to happen when when submitting. I thought it's best to leave it out.

src/cadenza.js
klinakerdisy: I had not taken into account that loading data in-memory might be an issue. Changed to make the browser handle the download.

Thanks for the proposal, it has been implemented.

src/cadenza.js
```diff
@@ -275,6 +279,72 @@ export class CadenzaClient {
...
+    const res = await fetch(url, { signal });
+    if (!res.ok) {
+      const message = 'failed to fetch';
+      switch (res.status) {
+        case 400:
+          throw new CadenzaError('bad-request', message);
+        case 401:
+          throw new CadenzaError('unauthorized', message);
+        case 404:
+          throw new CadenzaError('not-found', message);
```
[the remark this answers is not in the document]
klinakerdisy: Done

src/cadenza.js
```diff
@@ -502,6 +572,42 @@ function validGeometryType(value) {
   ].includes(value);
 }
 
+/** @param {string} [value] */
+function validMediaType(value) {
+  return value === CSV || value === MS_EXCEL_2007;
+}
+
+/** @param {Promise<Response>} responsePromise
```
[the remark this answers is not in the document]
klinakerdisy: Fixed

src/cadenza.js
klinakerdisy: Fixed

src/cadenza.js
klinakerdisy: This code is no longer needed because it's handled by the server.

PR #2, opened 2023-09-15
Tsm target origin

tsmoljo:


src/cadenza.js
```diff
@@ -92,9 +102,10 @@ export class CadenzaClient {
    * @param {string} baseUrl
    * @param {object} [options]
    * @param {HTMLIFrameElement | string} [options.iframe]
+   * @param {ExternalLinkKey} [options.targetOrigin]
    * @param {boolean} [options.debug]
    */
-  constructor(baseUrl, { debug = false, iframe } = {}) {
+  constructor(baseUrl, { debug = false, iframe, targetOrigin } = {}) {
```
jkissel, line 108: Maybe validate the new param using validRepositoryName and validWorkbookId?

src/cadenza.test.ts
```diff
@@ -152,6 +158,29 @@ describe('Given a Cadenza JS client instance', () => {
...
+      iframe = document.createElement('iframe');
+      cad = cadenza(BASE_URL, { iframe, targetOrigin: EXTERNAL_LINK_KEY });
+    });
+
+    it('Includes given origin param', () => {
+      cad.show(EMBEDDING_TARGET_ID);
+      const expectedKey = `webApplicationLink=${encodeURIComponent(
+        EXTERNAL_LINK_KEY.externalLinkId,
+      )}`;
+      const expectedKeyRepository = `webApplicationLinkRepository=${encodeURIComponent(
```
jkissel, line 175: The encoding should not be necessary for the test IDs, right? Proposal: Just interpolate the variables in the URL-string below, so that it's more clear how the expected URL looks like.

jkissel (changes requested):
Maybe add an example on the docs' start page with the new params? I guess having different origins is the default case.

jkissel (approved):

PR #3, opened 2023-09-19
add minScale to sandbox.html

ftarrach:
while looking over the DR I noticed that a control for the "minimum scale" was still missing


jkissel (approved):
Thx for this sandbox improvement :+1:

PR #5, opened 2023-09-27
client for JasperReport PDF downloading and embedding

Tobias43:


src/cadenza.js
```diff
@@ -151,12 +151,15 @@ export class CadenzaClient {
    * @return {Promise<void>} A Promise for when the iframe is loaded
    * @throws For an invalid source
    */
-  show(source, { hideMainHeaderAndFooter, hideWorkbookToolBar, signal, accept } = {}) {
+  show(
+    source,
+    { hideMainHeaderAndFooter, hideWorkbookToolBar, signal, accept } = {},
```
jkissel, line 156: The parameter for download and fetch was renamed to mediaType, so that it's the same as the actual query parameter. (Don't know why it was different in the first place ...) Please align it here as well.

src/cadenza.js
```diff
@@ -145,15 +145,21 @@ export class CadenzaClient {
...
    * @return {Promise<void>} A Promise for when the iframe is loaded
    * @throws For an invalid source
    */
-  show(source, { hideMainHeaderAndFooter, hideWorkbookToolBar, signal } = {}) {
+  show(
+    source,
+    { hideMainHeaderAndFooter, hideWorkbookToolBar, signal, accept } = {},
+  ) {
     this.#log('CadenzaClient#show', source);
     const params = new URLSearchParams({
```
jkissel, line 159: Proposal: Change the validMediaType() util to suppotedMediaType() and check the supported media types where applicable.

```
diff --git a/src/cadenza.js b/src/cadenza.js
index d777e23..346ae61 100644
--- a/src/cadenza.js
+++ b/src/cadenza.js
@@ -401,7 +401,7 @@ export class CadenzaClient {
    * Fetch data from a workbook view.
    *
    * @param {WorkbookViewSource} source - The workbook view to fetch data from
-   * @param {string} mediaType - The media type to use for the data
+   * @param {MediaType} mediaType - The media type to use for the data
    * @param {object} options - Options
    * @param {AbortSignal} [options.signal] - A signal to abort the data fetching
    * @return {Promise<Response>} A Promise for the fetch response
@@ -409,7 +409,10 @@ export class CadenzaClient {
    */
   fetchData(source, mediaType, { signal } = {}) {
     this.#log('CadenzaClient#fetchData', source, mediaType);
-    assert(validMediaType(mediaType), `Invalid media type: ${mediaType}`);
+    assert(
+      supportedMediaType(mediaType, [MediaType.CSV, MediaType.EXCEL]),
+      `Invalid media type: ${mediaType}`,
+    );
     const params = new URLSearchParams({ mediaType });
     return this.#fetch(resolvePath(source), { params, signal });
   }
@@ -442,14 +445,17 @@ export class CadenzaClient {
    * _Note:_ The file name, if not provided, is generated from the name of the workbook view and the current date.
    *
    * @param {WorkbookViewSource} source - The workbook view to download data from
-   * @param {string} mediaType - The media type to use for the data
+   * @param {MediaType} mediaType - The media type to use for the data
    * @param {object} options - Options
    * @param {string} [options.fileName] - The file name to use; The file extension is appended by Cadenza.
    * @throws For an invalid workbook view source or media type
    */
   downloadData(source, mediaType, { fileName }) {
     this.#log('CadenzaClient#downloadData', source, mediaType);
-    assert(validMediaType(mediaType), `Invalid media type: ${mediaType}`);
+    assert(
+      supportedMediaType(mediaType, [MediaType.CSV, MediaType.EXCEL]),
+      `Invalid media type: ${mediaType}`,
+    );
     const params = new URLSearchParams({
       mediaType,
       ...(fileName && { fileName }),
@@ -586,12 +592,23 @@ function validGeometryType(value) {
   ].includes(value);
 }
 
-/** @param {string} [value] */
-function validMediaType(value) {
-  const CSV = 'text/csv';
-  const MS_EXCEL_2007 =
-    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
-  return value === CSV || value === MS_EXCEL_2007;
+/**
+ * @typedef {string} MediaType - A media type
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types}
+ */
+
+const MediaType = /** @type {const Record<string, MediaType} */ {
+  CSV: 'text/csv',
+  EXCEL: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+  PDF: 'application/pdf',
+};
+
+/**
+ * @param {MediaType} type
+ * @param {MediaType[]} supportedTypes
+ */
+function supportedMediaType(type, supportedTypes) {
+  return supportedTypes.includes(type);
 }
 
 /**
```

README.md
```diff
@@ -69,6 +69,17 @@ cadenzaClient.show('{embeddingTargetId}', {
 - If the embedding target cannot be resolved, a 404 page is shown to the user.
 - Cadenza JS does not handle user authentication: If the user is not already logged in, the normal authentication flow of Cadenza will run. By default, the login page would be shown to the user.
 
+#### Show the Generated PDF of a Jasper Report View Directly
+Views of type "JasperReports report" can be shown in an iFrame like any other view.
```
jkissel, line 73: Isn't "JasperReports report" a bit redundant? -> "Jasper Report"

sandbox.html
```diff
@@ -159,6 +160,10 @@
     <label for="embeddingTargetId">Embedding target ID *</label>
     <input name="embeddingTargetId" id="embeddingTargetId" required>
   </div>
+  <div>
+    <label for="mediaType">MediaType (optional, for Jasper Reports)</label>
+    <input name="mediaType" id="mediaType" placeholder="application/pdf">
```
jkissel, line 165: Imo it doesn't make sense to use a specific value as the placeholder. In this case (since only PDF supported), I'd use a checkbox "Show Jasper Report as PDF".

src/cadenza.js
```diff
@@ -145,15 +145,21 @@ export class CadenzaClient {
    * @param {object} [options]
    * @param {boolean} [options.hideMainHeaderAndFooter] - Whether to hide the main Cadenza header and footer.
    * @param {boolean} [options.hideWorkbookToolBar] - Whether to hide the workbook toolbar.
+   * @param {string} [options.accept] - Media type; set to "application/pdf" for Jasper Report views
+   *     to show the PDF directly, without any Cadenza chrome around it.
```
jkissel, line 149: Although "chrome" is technically correct term, it always causes questions. I'd go with the words from the example "without any Cadenza footers and headers".

src/cadenza.js
```diff
@@ -442,7 +448,10 @@ export class CadenzaClient {
    * _Note:_ The file name, if not provided, is generated from the name of the workbook view and the current date.
    *
    * @param {WorkbookViewSource} source - The workbook view to download data from
-   * @param {string} mediaType - The media type to use for the data
+   * @param {string} mediaType - The media type to use for the data. Allowed are
+   * * 'text/csv'
+   * * 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+   * * 'application/pdf' (for Jasper Report views)
```
jkissel, line 454: Good to mention them here :+1:

Proposal: Put "(Excel)" behind the long one.

src/cadenza.js
```diff
@@ -591,7 +602,8 @@ function validMediaType(value) {
   const CSV = 'text/csv';
   const MS_EXCEL_2007 =
     'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
-  return value === CSV || value === MS_EXCEL_2007;
+  const PDF = 'application/pdf';
+  return !!value && [CSV, MS_EXCEL_2007, PDF].includes(value);
```
jkissel, line 606: For fetch/downloadData the mediaType is required, i.e. !!value is not correct.

src/cadenza.js
```diff
@@ -466,6 +475,8 @@ export class CadenzaClient {
     const url = this.#createUrl(path, params);
     const a = document.createElement('a');
     a.href = url.toString();
+    // causes the file to be downloaded even if the server sends a "Content-disposition: inline" header
+    a.download = '';
```
jkissel, line 479: :+1:

jkissel (changes requested):

README.md
Tobias43: To me "JasperReports" is just a product name, but I'm fine with your version, too. I'll change it.

src/cadenza.js
Tobias43: I had overlooked the rename, I'll change it.

sandbox.html
Tobias43: Sounds good, I'll change it.

src/cadenza.js
Tobias43: Right, that's less ambiguous :)

src/cadenza.js
Tobias43: Good idea, will do.

src/cadenza.js
Tobias43: That was to fix some type errors, but you're right, I better fix the types in the JSDoc instead.

src/cadenza.js
Tobias43: Your patch does not allow PDF in downloadData(). I'll add that. Shall we also allow it in fetchData()? It is not explicitly required in the story, but I can't think of a reason for not allowing it.

Tobias43: supportedMediaType() does not do much and is always used in combination with an assert. Shall we maybe put the assert insider the function and call it "assertMediaTypeSupported"?

README.md
Tobias43: Now I remember why I wrote it like that. "JasperReports report" is what the visualization type is called in Cadenza.

src/cadenza.js
jkissel: > Your patch does not allow PDF in downloadData(). I'll add that. Shall we also allow it in fetchData()? It is not explicitly required in the story, but I can't think of a reason for not allowing it.

Yes, please :+1: 

> supportedMediaType() does not do much and is always used in combination with an assert. Shall we maybe put the assert insider the function and call it "assertMediaTypeSupported"?

Sounds good :+1: I'd name it assertSupportedMediaType, though, so that the name is similar to the valid* functions.

README.md
jkissel: OK :shrug:

sandbox.html
```diff
@@ -70,11 +70,11 @@
 
     const cadenzaClient = cadenza(location.origin + '/trunk', { iframe: 'iframe', debug: true });
     const actionHandlers = {
-      show ({ embeddingTargetId, hideMainHeaderAndFooter, hideWorkbookToolBar, mediaType }) {
+      show ({ embeddingTargetId, hideMainHeaderAndFooter, hideWorkbookToolBar, jasperReportAsPdf }) {
         cadenzaClient.show(embeddingTargetId, {
           hideMainHeaderAndFooter: (hideMainHeaderAndFooter === 'on'),
           hideWorkbookToolBar: (hideWorkbookToolBar === 'on'),
-          accept: mediaType
+          ...(jasperReportAsPdf && { mediaType: 'application/pdf' })
```
jkissel, line 77: My comment was only about the sandbox. Here, I'd stick to the more future-proof mediaType, because this param is published is directly related to the endpoint param.

(Also the example diverges from this API atm.)

jkissel (changes requested):

sandbox.html
Tobias43: What do you mean with "stick to the more future-proof mediaType". I changed the option name from "accept" to "mediaType" as suggested in this comment: https://github.com/DisyInformationssysteme/cadenza.js/pull/5#discussion_r1338301472

And I don't find the error in the example in README.md. The only difference to here that I see is, that in the example I left out the optional options "hideMainHeaderAndFooter" and "hideWorkbookToolBar".

jkissel: Sry, my bad, we _are_ in sandbox code here 🤦

Still, you need to check for ==='on', otherwise the mediaType will always be set.

sandbox.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" xmlns="http://www.w3.org/1999/html">
```
jkissel, line 2: Why this ancient NS?

sandbox.html
Tobias43: Turns out that jasperReportAsPdf is undefined when the checkbox is not checked. So it worked correctly. But it is still nicer to check for "on" explicitly, so I changed it.

sandbox.html
Tobias43: No idea how that got in there, I definitely did not type it. Maybe IntelliJ generated it. Anyway, it's gone.

jkissel (approved):

PR #7, opened 2023-10-03
CADENZA-33493 [Embedding FA] Export Table and Indicator as JSON

anistor:
https://jira.disy.net/browse/CADENZA-33493

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ That's why the events use the `editGeometry` prefix.
 
 <small>API: [CadenzaClient#fetchData](./classes/CadenzaClient.html#fetchData)</small>
 
-Fetch data from a workbook view in CSV format.
+Fetch data from a workbook view in CSV or JSON format.
 
 ```javascript
 const response = await cadenzaClient.fetchData('{embeddingTargetId}', 'text/csv');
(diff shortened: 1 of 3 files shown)


Tobias43 (approved):

README.md
```diff
@@ -211,6 +211,14 @@ const text = await response.text();
 ...
 ```
 
+Fetch data from a workbook view in JSON format.
```
jkissel, line 214: We do not need to showcase every combination of method / options / value in these examples. For getting all the details, the apidoc is the better tool. That's what I tried to explain in the orange box here: https://conf.disy.net/pages/viewpage.action?spaceKey=EHB&title=Cadenza+JS#CadenzaJS-ExamplesintheREADME

So for me, another media type is not important enough to "justify" another two examples. What do you think?

sandbox.html
```diff
@@ -281,13 +288,34 @@
   </div>
   <div>
     <label for="mediaType">Media type *</label>
-    <input name="mediaType" id="mediaType" value="text/csv" required>
+    <select name="mediaType" id="mediaType" required>
+      <option>text/csv</option>
+      <option>application/vnd.openxmlformats-officedocument.spreadsheetml.sheet</option>
+      <option>application/json</option>
+      <option>application/pdf</option>
+    </select>
```
jkissel, line 296: :+1:

src/cadenza.js
```diff
@@ -642,6 +644,7 @@ function validGeometryType(value) {
 const MediaType = /** @type {Record<string, MediaType>} */ {
   CSV: 'text/csv',
   EXCEL: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+  JSON: 'application/json',
```
jkissel, line 647: Thx for maintaining the alphabetic order :+1:

README.md
anistor: Agreed, one media type showcased with a code snippet is enough.

PR #8, opened 2023-10-23
CADENZA-33816 [Workbook Embedding] Show welcome page and expand navig…

anistor:
…ator

https://jira.disy.net/browse/CADENZA-33816

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,40 @@ cadenzaClient.on('editGeometry:ok', (event) => {
 _Note:_ Under the hood, creating a geometry is similar to editing a geometry.
 That's why the events use the `editGeometry` prefix.
 
+### Highlight an Item in the Navigator
+
+<small>API: [CadenzaClient#show](./classes/CadenzaClient.html#show)</small>
+
+Show an embedding target in an iframe and highlight an item in the navigator. Additionally, expand the navigator tree.
+
+```javascript
+cadenzaClient.show({embeddingTargetId}, {expandNavigator: true, highlightGlobalId: 'ROOT.MyFolder'});
+```
+
+#### Highlight an Item in the Navigator on the Welcome Page
+
+<small>API: [CadenzaClient#show](./classes/CadenzaClient.html#show)</small>
+
+Show Cadenza's welcome page in an iframe and highlight an item in the navigator.
+
+```javascript
+cadenzaClient.show({page: 'welcome'}, {highlightGlobalId: 'ROOT.MyFolder'});
+```
+
+### Expand/collapse the navigator tree
+
+<small>API: [CadenzaClient#expandNavigator](./classes/CadenzaClient.html#expandNavigator)</small>
+
+Expand the navigator tree.
+
+```javascript
+const button = document.createElement('button');
+button.textContent = 'Expand Navigator';
+button.onclick = (event) => {
+  cadenzaClient.expandNavigator(true);
+};
+```
+
 ### Fetch Data From a Workbook View
 
 <small>API: [CadenzaClient#fetchData](./classes/CadenzaClient.html#fetchData)</small>
--- a/sandbox.html
+++ b/sandbox.html
@@ -71,8 +71,9 @@
     const contextPath = new URLSearchParams(location.search).get('contextPath') ?? '';
     const cadenzaClient = cadenza(location.origin + contextPath, { iframe: 'iframe', debug: true });
     const actionHandlers = {
-      show ({ embeddingTargetId, hideMainHeaderAndFooter, hideWorkbookToolBar, jasperReportAsPdf, highlightGlobalId, operationMode, disabledUiFeatures }) {
+      show ({ embeddingTargetId, expandNavigator, hideMainHeaderAndFooter, hideWorkbookToolBar, jasperReportAsPdf, highlightGlobalId, operationMode, disabledUiFeatures }) {
         cadenzaClient.show(embeddingTargetId, {
+          expandNavigator: (expandNavigator === 'on'),
           hideMainHeaderAndFooter: (hideMainHeaderAndFooter === 'on'),
           hideWorkbookToolBar: (hideWorkbookToolBar === 'on'),
           highlightGlobalId,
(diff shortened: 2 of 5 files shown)


src/cadenza.js
```diff
@@ -201,6 +209,7 @@ export class CadenzaClient {
    * @param {Geometry} [options.geometry] - A geometry to show on the map
    * @param {boolean} [options.hideMainHeaderAndFooter] - Whether to hide the main Cadenza header and footer
    * @param {boolean} [options.hideWorkbookToolBar] - Whether to hide the workbook toolbar
+   * @param {boolean} [options.expandNavigator] - Indicates if the navigator should be expanded.
    * @param {GlobalId} [options.highlightGlobalId] - The ID of an item to highlight / expand in the navigator
```
klschm, line 213: Here exactly is what I thought: The highlightGlobalId is exactly expanding the navigator. So do we really need both parameters? And if yes, why? What's the example use case for it?

sandbox.html
```diff
@@ -70,14 +70,24 @@
...
           hideMainHeaderAndFooter: (hideMainHeaderAndFooter === 'on'),
           hideWorkbookToolBar: (hideWorkbookToolBar === 'on'),
           highlightGlobalId,
           ...(jasperReportAsPdf === 'on' && { mediaType: 'application/pdf' }),
         });
       },
+      showPage ({ page, expandNavigator, highlightGlobalId }) {
+        cadenzaClient.show({ page: page }, {
+          expandNavigator: (expandNavigator === 'on'),
+          highlightGlobalId
```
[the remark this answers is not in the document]
anistor: These are 2 different things, but closely related, hence the possible confusion. As the name implies, `expandNavigator` expands the left navigation panel, making the navigation tree visible. 

highlightGlobalId selects the given item from the tree and also expands the tree item (if it can be expanded). But has nothing to do with the tree panel being visible.

src/cadenza.js
anistor: no, highlightGlobalId does not expand the navigator, it just selects and expands the tree item.

sandbox.html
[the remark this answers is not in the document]
anistor: AFAIK, _this_ is the discussion. that's why we do a review :)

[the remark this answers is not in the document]
klschm: I don't know whether or not this is the discussion for naming. But I will clearify that with @jkissel . For naming I would propose "showNavigator" instead of "expandNavigator".

src/cadenza.js
```diff
@@ -54,6 +54,11 @@ globalThis.cadenza = Object.assign(
 /** @typedef {EmbeddingTargetId | WorksheetKey} WorksheetSource - A worksheet source */
 /** @typedef {EmbeddingTargetId | WorkbookViewKey} WorkbookViewSource - A workbook view source */
 
+/**
+ * @typedef PageSource - A Cadenza landing page
```
jkissel, line 58: What distinguishes a page from a "landing page"? Proposal: "A well-known Cadenza page"

src/cadenza.js
```diff
@@ -54,6 +54,11 @@ globalThis.cadenza = Object.assign(
...
+ * @property {string} page - The name of the landing page
```
jkissel, line 59: string -> 'welcome'

src/cadenza.test.ts
```diff
@@ -188,6 +193,17 @@ describe('Given a Cadenza JS client instance', () => {
     });
   });
 
+  describe('When showing the welcome page', () => {
```
jkissel, line 196: Thx for extending the tests :+1:

jkissel (changes requested):

sandbox.html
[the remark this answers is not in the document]
anistor: I think `showNavigator()` would indeed conflict a bit with what we're used from `show()` / `showMap()` so I'd propose to stick with `expandNavigator()`.

src/cadenza.js
anistor: I do not understand what you mean by this.

README.md
```diff
@@ -198,6 +211,26 @@ cadenzaClient.on('editGeometry:ok', (event) => {
...
+button1.onclick = (event) => {
+  cadenzaClient.expandNavigator(true);
+};
+
+const button2 = document.createElement('button');
+button2.textContent = 'Collapse Navigator';
+button2.onclick = (event) => {
+  cadenzaClient.expandNavigator(false);
+};
+```
```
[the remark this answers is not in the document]
anistor: Ok, I'll remove button2 sample used for collapsing the navigator.

sandbox.html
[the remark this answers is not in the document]
jkissel: OK :+1:

jkissel (changes requested):

sandbox.html
```diff
@@ -208,6 +226,35 @@
...
+  <div>
+    <label for="page">Page *</label>
+    <select name="page" id="page" required>
+      <option>welcome</option>
+      <option>an-invalid-page-name</option>
+    </select>
+  </div>
+  <div>
+    <label>
+      <input type="checkbox" name="expandNavigator">
```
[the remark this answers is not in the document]
anistor: Yes, but let's not think about welcome page only. In theory there might be others.

src/cadenza.js
anistor: Oh, got it now. But I really do not think it should be the compiler error. It should be a runtime error, as it happens for all other arg validations we have in the library.

Also, why directly limit this to 'welcome'? In theory the api should support other pages.

sandbox.html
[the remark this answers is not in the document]
jkissel: :+1:

src/cadenza.js
jkissel: > It should be a runtime error
Why would you let people call this method if we know at build time that the argument would be wrong? That's the point of TS, right?

> as it happens for all other arg validations we have in the library.
The validation for the other cases are also on top of maximum specific types. The validations add additional safety for when we cannot express certain things as types or for users that do not use TS.

> Also, why directly limit this to 'welcome'? In theory the api should support other pages.
If we would add another well-known page, we'd also extend the validation.

anistor: That is the point of TS indeed but this is JS, so we're still doing things like validateGeometryType. And we even have unit tests for these validations, so they are not expected to be compile errors, 'cause that would not even allow us to write a unit test.

If I do the suggested change I get:
{code}
    src/cadenza.test.ts:198:16 - error TS2345: Argument of type '{ page: string; }' is not assignable to parameter of type 'string | PageSource | WorkbookKey | WorksheetKey | WorkbookViewKey'.
      Type '{ page: string; }' is not assignable to type 'PageSource'.
        Types of property 'page' are incompatible.
          Type 'string' is not assignable to type '"welcome"'.

    198       cad.show(WELCOME_PAGE);
                       ~~~~~~~~~~~~
{code}

Why should we insist doing for page source something differently that has been done for media type and geometry type?

jkissel: For MediaType and GeometryType we also have @typedefs. Based on these typedefs we generate a .d.ts file for people who use typescript.

anistor: I fixed the previous compilation error by importing the PageSource type, so we should be good to go.

jkissel (approved):

anistor:
Thanks @jkissel , @klschm !

PR #9, opened 2023-10-27
Introduce support for new parameters

klinakerdisy:
"operationMode" and "disabledUiFeatures".

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,26 @@ cadenzaClient.show('{embeddingTargetId}', {
 });
 ```
 
+#### Show an Embedding Target in Simplified Operation Mode
+
+Workbook embedding targets can be shown with simplified operation mode enabled by setting the "operationMode" option to "simplified".
+
+```javascript
+cadenzaClient.show('{embeddingTargetId}', {
+  operationMode: 'simplified'
+});
+```
+
+#### Show an Embedding Target with Disabled Features
+
+Embedding targets can be shown with certain UI features disabled. To disable the designer, add "workbook-design" to the "disabledUiFeatures" option. To disable editing of workbook layout and design, add "workbook-view-management".
+
+```javascript
+cadenzaClient.show('{embeddingTargetId}', {
+  disabledUiFeatures: ['workbook-design', 'workbook-view-management']
+});
+```
+
 ### Abort (Iframe) Loading
 
 Cadenza JS uses the [AbortController Web API](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for aborting requests. This is supported by most of the public methods.
--- a/README.md
+++ b/README.md
@@ -65,12 +65,14 @@ const cadenza = window.cadenza.noConflict();
 
 <small>API: [CadenzaClient#show](./classes/CadenzaClient.html#show)</small>
 
-Show an embedding target in an iframe and hide Cadenza's main header and footer as well as the workbook toolbar:
+Show an embedding target in an iframe and hide Cadenza's main header and footer as well as the workbook toolbar. Additionally, enable simplified operation mode and disable the designer.
 
 ```javascript
 cadenzaClient.show('{embeddingTargetId}', {
   hideMainHeaderAndFooter: true,
   hideWorkbookToolBar: true,
+  operationMode: 'simplified',
+  disabledUiFeatures: ['workbook-design']
 });
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -87,6 +87,26 @@ cadenzaClient.show('{embeddingTargetId}', {
...
+#### Show an Embedding Target with Disabled Features
+
+Embedding targets can be shown with certain UI features disabled. To disable the designer, add "workbook-design" to the "disabledUiFeatures" option. To disable editing of workbook layout and design, add "workbook-view-management".
+
+```javascript
+cadenzaClient.show('{embeddingTargetId}', {
+  disabledUiFeatures: ['workbook-design', 'workbook-view-management']
+});
+```
+
```
jkissel, line 109: To keep the list of examples concise, maybe it would be enough to add these params to the existing "Show an embedding target" example?

src/cadenza.js
```diff
@@ -64,6 +64,13 @@ globalThis.cadenza = Object.assign(
  * _Note:_ The GeoJSON geometry type "GeometryCollection" is currently not supported.
  */
 /** @typedef {[number,number,number,number]} Extent - An array of numbers representing an extent: [minx, miny, maxx, maxy] */
+/**
+ * @typedef {[string]} DisabledUiFeatures - An array of strings representing Cadenza UI features to disable
+ *
+ * _Note:_ Supported features are:
+ * * 'workbook-design' - Disable the designer
+ * * 'workbook-view-management' - Disable workbook layout/design editing
+ * */
```
jkissel, line 73: To me it would be more sensible have UiFeature as the type and then disabledUiFeatures would just be an array of those.

```javascript
/**
 * @typedef {'workbook-design' | 'workbook-view-management'} UiFeature - The name of a Cadenza UI feature
 *
 * Supported features are:
 * - `workbook-design`: The designer
 * - `workbook-view-management` - Add/Edit/Remove workbook views
 */
```
```javascript
   * @param {UiFeature[]} [options.disabledUiFeatures] - Cadenza UI features to disable
```

(Btw, an array of strings would be string[]. [string] is an array with exactly one string value.)

src/cadenza.js
```diff
@@ -165,6 +172,8 @@ export class CadenzaClient {
    * @param {GlobalId} [options.highlightGlobalId] - The ID of an item to highlight / expand in the navigator
    * @param {string} [options.mediaType] - Set to 'application/pdf' for views of type "JasperReports report"
    *     to show the report PDF directly, without any Cadenza headers or footers.
+   * @param {string} [options.operationMode] - Whether to enable simplified operation mode
```
jkissel, line 175: Proposal: Use a type for the operation mode, too.
```javascript
/** @typedef {'normal' | 'simplified'} WorkbookOperationMode - The mode in which a workbook should be operated */
```

jkissel (changes requested):

README.md
klinakerdisy: How's that look?

It didn't make sense to include 'workbook-view-management' in the example, because it has no effect in simplified operation mode, and nobody is going to specify "operationMode: 'normal'" since that's the default anyway.

src/cadenza.js
klinakerdisy: Good idea.

Oops, didn't quite pay attention there.

src/cadenza.js
klinakerdisy: Done (y)

src/cadenza.js
```diff
@@ -64,6 +64,14 @@ globalThis.cadenza = Object.assign(
  * _Note:_ The GeoJSON geometry type "GeometryCollection" is currently not supported.
  */
 /** @typedef {[number,number,number,number]} Extent - An array of numbers representing an extent: [minx, miny, maxx, maxy] */
+/** @typedef {'normal'|'simplified'} OperationMode - The mode in which a workbook should be operated */
+/**
+ * @typedef {'workbook-design'|'workbook-view-management'} UiFeature - The name of a Cadenza UI feature
+ *
+ * _Note:_ Supported features are:
+ * * 'workbook-design' - Disable the designer
+ * * 'workbook-view-management' - Disable workbook layout/design editing
```
jkissel, line 73: Maybe add the hint from the sprint review? "Is included in 'workbook-design'."

jkissel (approved):

src/cadenza.js
klinakerdisy: Added :)

PR #10, opened 2023-10-31
Slb/an/cadenza 33417

anistor:


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mediaType` option for `CadenzaClient#show`
 - `CadenzaClient#fetchData`
 - `CadenzaClient#downloadData`
+- `parts` option for `CadenzaClient#fetchData` and `CadenzaClient#downloadData`
```
jkissel, line 16: Since the whole #fetchData and #downloadData methods are new, imo it's not necessary to have an extra line in the changelog for this new option.

src/cadenza.js
```diff
@@ -163,7 +163,7 @@ export class CadenzaClient {
    * @param {boolean} [options.hideMainHeaderAndFooter] - Whether to hide the main Cadenza header and footer
    * @param {boolean} [options.hideWorkbookToolBar] - Whether to hide the workbook toolbar
    * @param {GlobalId} [options.highlightGlobalId] - The ID of an item to highlight / expand in the navigator
-   * @param {string} [options.mediaType] - Set to 'application/pdf' for views of type "JasperReports report"
+   * @param {MediaType} [options.mediaType] - Set to 'application/pdf' for views of type "JasperReports report"
```
jkissel, line 166: The for this fix :+1:

src/cadenza.js
```diff
@@ -436,14 +436,18 @@ export class CadenzaClient {
    * * 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' (Excel)
    * * 'text/csv'
    * @param {object} options - Options
+   * @param {TableParts[]} [options.parts] - An optional array of table parts to export
```
jkissel, line 439: Imo the type name should be singular (like MediaType).

jkissel, line 439: "If not given, all parts are exported."

src/cadenza.js
```diff
@@ -608,6 +616,25 @@ function validGeometryType(/** @type string */ value) {
...
+  TOTALS: 'totals',
+};
+
+function assertSupportedTableParts(/** @type TableParts[] */ parts) {
+  assert(Array.isArray(parts), `'parts' parameter must be an array`);
+  assert(
+    parts.every((p) => Object.values(TableParts).includes(p)),
+    `Invalid 'parts' parameter value: ${parts}`,
+  );
+}
```
jkissel, line 636: Proposal: Make the values part of the type definition instead of only mentioning them in the docs.
```typescript
/** @typedef {'columns' | 'values' | 'totals'} TablePart - A part of a table to export */

function assertValidTablePart(/** @type TablePart */ value) {
  assert(
    [ 'columns', 'values', 'totals' ].includes(value),
    `Invalid table part: ${value}`);
}

...

parts.forEach(assertValidTablePart);
```

src/cadenza.js
```diff
@@ -638,7 +665,8 @@ function assertSupportedMediaType(
  * @param {GlobalId} [params.highlightGlobalId]
  * @param {string} [params.locationFinder]
  * @param {Extent} [params.mapExtent]
- * @param {string} [params.mediaType]
+ * @param {MediaType} [params.mediaType]
+ * @param {TableParts[]} [params.parts]
```
jkissel, line 669: Please move the new param to the correct place in the alphabetic ordering.

src/cadenza.js
```diff
@@ -671,6 +700,7 @@ function createParams({
     ...(locationFinder && { locationFinder }),
     ...(mapExtent && { mapExtent: mapExtent.join() }),
     ...(mediaType && { mediaType }),
+    ...(parts && { parts: parts.toString() }),
```
jkissel, line 703: I'd rather always use the same method of joining an array, i.e. .join() (like above).

jkissel (changes requested):

anistor:
I've addressed all comments added up to now. Thanks @jkissel !

CHANGELOG.md
anistor: done

src/cadenza.js
```diff
@@ -436,14 +436,18 @@ export class CadenzaClient {
...
    * @param {AbortSignal} [options.signal] - A signal to abort the data fetching
    * @return {Promise<Response>} A Promise for the fetch response
    * @throws For an invalid workbook view source or media type
    */
-  fetchData(source, mediaType, { signal } = {}) {
+  fetchData(source, mediaType, { parts, signal } = {}) {
     this.#log('CadenzaClient#fetchData', source, mediaType);
     assertSupportedMediaType(mediaType);
-    const params = createParams({ mediaType });
+    if (parts) {
```
jkissel, line 447: Proposal: Move the if() to the assertValidTableParts to have it only once.

src/cadenza.js
```diff
@@ -608,6 +616,26 @@ function validGeometryType(/** @type string */ value) {
   ].includes(value);
 }
 
+/** @typedef {'columns' | 'values' | 'totals'} TablePart - A part of a table to export. */
+
+const TablePart = /** @type {Record<string, TablePart>} */ {
```
jkissel, line 621: This mapping is unused in Cadenza JS, an array of the values would be sufficient.

jkissel (approved):
Looks good now (y)

src/cadenza.js
anistor: Should the client code be able to use those constants? (if our code does not use them directly)

I can remove them just making sure it is really intended to obliterate them.
I see for geometries we do not have defined individual constants just an array of the possible values (as you suggest here) but for MediaTypes we have constants (which we also use in our code directly). Which of the 2 approaches is best?

README.md
```diff
@@ -221,6 +221,18 @@ const text = await response.text();
...
+
+Fetch data from a workbook view in JSON format and include only the data values and the aggregations (totals) and exclude
+the column names.
+
+```javascript
+const response = await cadenzaClient.fetchData('{embeddingTargetId}', 'application/json', { parts: ['values', 'totals'] });
+
+const tableData = await response.json();
+...
+```
```
jkissel, line 234: Nice example :+1:

src/cadenza.js
jkissel: The geometry types are used inside of the lib. If clients should be able to use this mapping, it needs to be exported and should be marked as @readonly and really made readonly using Object.freeze().

jkissel (approved):

anistor:
Thanks @jkissel !

PR #13, opened 2024-01-27
add documentation for 'change:extent' event

jpalys:


src/cadenza.js
```diff
@@ -204,6 +204,8 @@ export class CadenzaClient {
    *   view, each row includes the geometry of the selected object as the last value.
    *   <p>
    *   See also: [JSON Representation of Cadenza Object Data](../index.html#md:json-representation-of-cadenza-object-data)
+   * @fires `change:extent` - When the map extent is changed a POST message 'change:extent' is fired.
+   *   The event details contain the new {@link Extent} of the map (transformed to EPSG:4326).
```
jkissel, line 208: On Friday, I improved the way how events are documented: https://github.com/DisyInformationssysteme/cadenza.js/commit/6020ae81bb42e257be3a87361f31d2708ef20b38

Goals:
- Remove the duplication of the documentation if the event is fired from multiple methods (like in your case).
- Document the type of the event detail.

Proposal for your event:
```javascript
/** @typedef {CadenzaEvent<'change:extent', {extent: Extent}>} CadenzaChangeExtentEvent - When the user moved the map. The extent is transformed according to the `useMapSrs` option. */
```

(Sry, I should've informed you about the change. I hope I didn't cause too much trouble.)

src/docs.md
```diff
@@ -205,6 +205,19 @@ const unsubscribe = cadenzaClient.on('editGeometry:ok', (event) => ...);
 unsubscribe();
 ```
 
+### Subscribe to map extent change
+
+Subscribe to `extent:change` event to be notified about map extent changes. The event details contain the new map extent (transformed to EPSG:4326):
```
jkissel, line 210: It's `change:extent`. But if I look at the example - maybe it's not really necessary, because the event is documented in the api docs and we already have an example for subscribing to an event in TS. What do you think?

If you keep the example (and resolved the above comment), you should use the `CadenzaChangeExtentEvent` type.

jkissel (changes requested):

PR #14, opened 2024-01-30
Slb/ab/set selection

AndreasWBartels:


sandbox.html
```diff
@@ -406,6 +408,15 @@
   </div>
 </template>
 
+<template data-action="setSelection">
+  <div>
+    <label for="layerPath">Path to the dataview layer * (GeoJSON Array)</label>
```
jkissel, line 413: * I think it should be "data view layer" to be proper English.
* It's not a GeoJSON array.

sandbox.html
```diff
@@ -406,6 +408,15 @@
...
+    <textarea name="layerPath" id="layerPath" rows="5" required></textarea>
+    <label for="values">List of Object identifiers * (GeoJSON Array)</label>
+    <textarea name="values" id="values" rows="5" required></textarea>
+  </div>
```
jkissel, line 417: There should be a <div> around every form control and its label. There is some spacing defined between these containers.

sandbox.html
```diff
@@ -406,6 +408,15 @@
   </div>
 </template>
 
+<template data-action="setSelection">
+  <div>
+    <label for="layerPath">Path to the dataview layer * (GeoJSON Array)</label>
+    <textarea name="layerPath" id="layerPath" rows="5" required></textarea>
+    <label for="values">List of Object identifiers * (GeoJSON Array)</label>
```
jkissel, line 415: Also not a GeoJSON array.

src/cadenza.js
```diff
@@ -323,6 +324,21 @@ export class CadenzaClient {
     return this.#postRequest('setFilter', { filter });
   }
 
+  /**
+   * Set filter Selection in the currently shown map view.
```
jkissel, line 328: It's not about filters -> ~filter~

src/cadenza.js
```diff
@@ -323,6 +324,21 @@ export class CadenzaClient {
...
+   *
+   * @param {WorkbookLayerPath} layerPath - The variable values
```
jkissel, line 330: Copy/paste mistake: This param is not for the values.

Proposal:
- Name the param just "layer", because the path is an implementation detail.
- Description: "The data view layer to set the selection in"

src/cadenza.js
```diff
@@ -323,6 +324,21 @@ export class CadenzaClient {
...
+   * @param {string[]} values - The variable values
+   * @return {Promise<void>} A `Promise` for when the filter variables were set.
```
jkissel, line 332: "A `Promise` for when the selection was set."

src/docs.md
```diff
@@ -92,6 +92,7 @@ cadenzaClient.show('{embeddingTargetId}', {
 });
 ```
 
+- The embedded target id can be defined for an open "workbook" in the menu under "more -> Manage workbook -> Embeddeding"
```
jkissel, line 95: Good hint :+1:

jkissel (changes requested):

PR #15, opened 2024-01-30
Extend "selectObjects" method with "selectableLayers" parameter to al…

klinakerdisy:
…low limiting selection to specific map layers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CadenzaClient#selectObjects`
 - `CadenzaChangeExtentEvent`
 - Documentation for the "JSON Representation of Cadenza Object Data"
+- `selectableLayers` parameter to `CadenzaClient#selectObjects` and relevant documentation 
 
 ### Changed
 - Improved the documentation and types of events.
--- a/sandbox.html
+++ b/sandbox.html
@@ -148,6 +148,7 @@
       mapExtent,
       minScale,
       parts,
+      selectableLayers,
       simplifiedOperationMode,
       useMapSrs
     }) {
@@ -167,6 +168,7 @@
         minScale,
         parts: parts && parts.split(','),
         ...(simplifiedOperationMode === 'on' && { operationMode: 'simplified' }),
+        selectableLayers: selectableLayers && JSON.parse(selectableLayers),
         useMapSrs: useMapSrs === 'on'
       };
     }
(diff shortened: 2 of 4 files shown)


src/cadenza.js
```diff
@@ -417,14 +418,15 @@ export class CadenzaClient {
...
-    { locationFinder, mapExtent, useMapSrs, signal } = {},
+    { locationFinder, mapExtent, useMapSrs, selectableLayers, signal } = {},
   ) {
     this.#log('CadenzaClient#selectObjects', backgroundMapView);
     const params = createParams({
       action: 'selectObjects',
       locationFinder,
       mapExtent,
       useMapSrs,
+      selectableLayers,
```
dalvarezmartinez1, line 429: Is this parameter already supported by the #show command?

dalvarezmartinez1 (commented):
done

src/cadenza.js
jkissel: I think it's relevant only for the selectObjects action.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CadenzaClient#selectObjects`
 - `CadenzaChangeExtentEvent`
 - Documentation for the "JSON Representation of Cadenza Object Data"
+- `selectableLayers` parameter to `CadenzaClient#selectObjects` and relevant documentation
```
jkissel, line 14: The whole method is new in this release (see three lines above), i.e. it's not necessary to mention this option specifically in the changelog.

sandbox.html
```diff
@@ -148,6 +148,7 @@
       mapExtent,
       minScale,
       parts,
+      selectableLayers,
```
jkissel, line 151: The story description looks to me as if there should be an option to define _a single_ layers (not multiple).
> Within the selection dialog, only features from specified layer are selected if layer name was provided.
If no layer name was given, the selection takes place in current algorithm.

Are you sure that it should be multiple?

Proposal: Name the option just "layer" in case we want to identify a layer for other reasons in the future.

jkissel (changes requested):

sandbox.html
tsmoljo: The description was faulty, I updated description and added some bullet points on expected behavior when multiple layers are specified

jkissel: OK, then it's all good :+1:

jkissel (approved):

src/docs.md
```diff
@@ -223,6 +223,20 @@ cadenzaClient.on('selectObjects:ok', (event) => {
...
 
+#### Limit selection to specific map layer
+
+<small>API: [CadenzaClient#selectObjects](./classes/CadenzaClient.html#selectObjects)</small>
+
+Limit object selection to specific map layers. For layers in groups, pass the layer path.
+
+```javascript
+cadenzaClient.selectObjects('{embeddingTargetId}', {
+  selectableLayers: [
```
jkissel, line 234: Here it's still selectedLayers.

klinakerdisy: Fixed. Thanks!

PR #17, opened 2024-02-09
Slb/ab/map layer object selection

AndreasWBartels:


src/cadenza.js
```diff
@@ -300,6 +300,7 @@ export class CadenzaClient {
       operationMode,
       useMapSrs,
     });
+    params.append('targetType', 'MAP');
```
jkissel, line 303: Why not passing this to createParams()?

src/cadenza.js
```diff
@@ -327,6 +328,51 @@ export class CadenzaClient {
     return this.#postRequest('setFilter', { filter });
   }
 
+  /**
+   * Set Selection in the currently shown map view.
+   *
+   * @param {WorkbookLayerPath} layer - The data view layer to set the selection in
+   * @param {string[]} values - The variable values
```
jkissel, line 335: It's actually `unknown[]`, right?

src/cadenza.js
```diff
@@ -327,6 +328,51 @@ export class CadenzaClient {
...
+   * @return {Promise<void>} A 'Promise' for when the selection was set.
+   */
+  setSelection(
+    /** @type WorkbookLayerPath */ layer,
+    /** @type string[] */ values,
```
jkissel, line 340: The `@type` annotations are redundant, because of the above `@param`s.

src/cadenza.js
```diff
@@ -327,6 +328,51 @@ export class CadenzaClient {
...
+   * @param {WorkbookLayerPath} layer - The data view layer to remove the selection from
+   * @param {string[]} values - The variable values
+   * @return {Promise<void>} A 'Promise' for when the selection was removed.
+   */
+  removeSelection(
+    /** @type WorkbookLayerPath */ layer,
+    /** @type string[] */ values,
+  ) {
+    this.#log('CadenzaClient#removeSelection', layer, values);
+    return this.#postRequest('removeSelection', { layer, values });
```
jkissel, line 373: Would it maybe make our live in Cadenza easier if we used only one event with a `mode?: 'add' | 'remove'`? (The API here is fine :+1:)

src/docs.md
```diff
@@ -92,6 +92,7 @@ cadenzaClient.show('{embeddingTargetId}', {
 });
 ```
 
+- The embedded target id can be defined for an open "workbook" in the menu under "more -> Manage workbook -> Embeddeding"
```
jkissel, line 95: Maybe rather add this description to the EmbeddingTargetId @typedef.

src/cadenza.js
```diff
@@ -327,6 +328,51 @@ export class CadenzaClient {
...
+  addSelection(
+    /** @type WorkbookLayerPath */ layer,
+    /** @type string[] */ values,
+  ) {
+    this.#log('CadenzaClient#addSelection', layer, values);
+    return this.#postRequest('addSelection', { layer, values });
+  }
+
+  /**
+   * Remove Selection in the currently shown map view.
```
jkissel, line 362: "Remove selection" could be confused with "Unset selection". Proposal: "Add to / Remove from the selection in the currently shown map view."

jkissel (changes requested):
Please also mention the new methods in the CHANGELOG and add an example to the docs.ts (after the showMap examples).

src/docs.md
jkissel: I did that meanwhile, because it was requested by solutioneers.

sandbox.html
```diff
@@ -457,6 +463,39 @@
   </div>
 </template>
 
+<template data-action="setSelection">
```
jkissel, line 466: You could define this template once and reuse it for the different actions, like it's done e.g. with "common-show".

PR #18, opened 2024-02-15
CADENZA-35393 [Cadenza API] Show and Hide Layers in Embedded Map

anistor:


src/cadenza.js
```diff
@@ -432,6 +432,19 @@ export class CadenzaClient {
...
+  /**
+   * Set the visibility of a layer in an embedded map view. The map view must be shown prior to calling this.
+   *
+   * @param {WorkbookLayerPath} layer - The layer to show or hide, identified by its path
+   * @param {boolean} visible - The visibility state of the layer
+   */
+  setLayerVisibility(layer, visible = true) {
+    this.#log('CadenzaClient#setLayerVisibility', ...arguments);
+    if (layer) {
+      this.#postEvent('setLayerVisibility', { layer, visible });
```
klinakerdisy, line 444: As you suggested we should use `#postRequest` here instead.

sandbox.html
```diff
@@ -457,6 +459,19 @@
...
 
+<template data-action="setLayerVisibility">
+  <div>
+    <label for="layer">Layer name *</label>
+    <input name="layer" id="layer" required>
+  </div>
+  <div>
+    <label>
+      <input type="checkbox" name="visibility">
+      Layer visibility
```
klinakerdisy, line 470: Maybe change this to "Visible" instead?

sandbox.html
```diff
@@ -136,6 +136,7 @@
       showPage: data => cadenzaClient.show({ page: data.page }, getOptions(data)),
       expandNavigator: data => cadenzaClient.expandNavigator(data.expandNavigator === 'on'),
       showMap: data => cadenzaClient.showMap(data.embeddingTargetId, getOptions(data)),
+      setLayerVisibility: data => cadenzaClient.setLayerVisibility(data.layer.split(','), data.visibility === 'on'),
```
jkissel, line 139: In the other place where we ask for a layer, we parse JSON. I think this should be done consistently and it should be clear for the user what to enter.

sandbox.html
```diff
@@ -457,6 +459,19 @@
   </div>
 </template>
 
+<template data-action="setLayerVisibility">
+  <div>
+    <label for="layer">Layer name *</label>
```
jkissel, line 464: It's a layer path. Proposal: Just "Layer"

src/cadenza.js
jkissel: Background: #postEvent is for sync, where we do not expect errors, #postRequest for async, where we need to know whether it worked. Since here we execute a workbook command in Cadenza, #postRequest is the right tool.

src/cadenza.js
```diff
@@ -432,6 +432,19 @@ export class CadenzaClient {
...
 
+  /**
+   * Set the visibility of a layer in an embedded map view. The map view must be shown prior to calling this.
+   *
+   * @param {WorkbookLayerPath} layer - The layer to show or hide, identified by its path
+   * @param {boolean} visible - The visibility state of the layer
+   */
+  setLayerVisibility(layer, visible = true) {
+    this.#log('CadenzaClient#setLayerVisibility', ...arguments);
+    if (layer) {
```
jkissel, line 443: This if() is not needed, the type is not nullable.

src/docs.md
```diff
@@ -230,6 +230,18 @@ cadenzaClient.on('selectObjects:ok', (event) => {
 });
 ```
 
+### Change visibility of Layers in a Workbook Map
+
+<small>API: [CadenzaClient#setLayerVisibility](./classes/CadenzaClient.html#setLayerVisibility)</small>
+
+To change the visibility of a layer in an already shown map view, pass the layer path and the desired visibility.
+For layers in groups, the layer path contains the ancestors also. When making a layer visible, its ancestors will
+be made visible also. When hiding a layer, the ancestors are not affected.
```
jkissel, line 239: Good hint :+1: Maybe rather add that to the API docs?

jkissel (changes requested):

sandbox.html
anistor: In that other place we ask for multiple layers. So, JSON here?

src/cadenza.js
anistor: ok

src/docs.md
anistor: ok

sandbox.html
jkissel: At least you should add a hint below the input (<small>), what the expectation is.

anistor: I tried making it JSON. But I can't seem to find the right json syntax for a single array (not a json object).

jkissel:
```
["groupName", "layerName"]
```
?

anistor: .... I was trying with single quotes, and that's not valid json of course :smile:  :brain: :gun:

klinakerdisy (approved):

anistor:
I hope I have addressed all comments by now. Please take a look again.

sandbox.html
```diff
@@ -346,6 +348,14 @@
...
     <label for="parts">Parts</label>
     <input name="parts" id="parts" placeholder="columns,values,totals">
+    <small>
+      A comma-separated list of parts. Supported parts are:
+      <ul>
+        <li>columns</li>
+        <li>values</li>
+        <li>totals</li>
+      </ul>
+    </small>
```
jkissel, line 358: Nice :+1:

src/cadenza.js
```diff
@@ -432,6 +432,20 @@ export class CadenzaClient {
...
+   * For layers in groups, the layer path contains the ancestors also. When making a layer visible, its ancestors will
+   * be made visible also. When hiding a layer, the ancestors are not affected. The operation is asynchronous and its
+   * completion is signaled by an event of type 'setLayerVisibility:success' or 'setLayerVisibility:error'.
+   *
+   * @param {WorkbookLayerPath} layer - The layer to show or hide, identified by its path
+   * @param {boolean} visible - The visibility state of the layer
+   */
+  setLayerVisibility(layer, visible = true) {
+    this.#log('CadenzaClient#setLayerVisibility', ...arguments);
+    this.#postRequest('setLayerVisibility', { layer, visible });
```
jkissel, line 446: You also need to return that Promise, otherwise the client can still not know whether it worked.

src/cadenza.js
```diff
@@ -432,6 +432,20 @@ export class CadenzaClient {
...
+  /**
+   * Set the visibility of a layer in an embedded map view. The map view must be shown prior to calling this.
+   * For layers in groups, the layer path contains the ancestors also. When making a layer visible, its ancestors will
+   * be made visible also. When hiding a layer, the ancestors are not affected. The operation is asynchronous and its
+   * completion is signaled by an event of type 'setLayerVisibility:success' or 'setLayerVisibility:error'.
+   *
+   * @param {WorkbookLayerPath} layer - The layer to show or hide, identified by its path
+   * @param {boolean} visible - The visibility state of the layer
+   */
+  setLayerVisibility(layer, visible = true) {
```
jkissel, line 444: I find a setter with a default for the value surprising. If you want to keep it, then you need to make the param optional in the JSDoc:
```
@param {WorkbookLayerPath} [layer=true] layer - ...
```
(Here the JSDoc is used for the TS types.)

src/docs.md
```diff
@@ -230,6 +230,21 @@ cadenzaClient.on('selectObjects:ok', (event) => {
...
+
+<small>API: [CadenzaClient#setLayerVisibility](./classes/CadenzaClient.html#setLayerVisibility)</small>
+
+To change the visibility of a layer in an already shown map view, pass the layer path and the desired visibility.
+The operation is asynchronous and completion is signaled with an event. 
+
+```javascript
+cadenzaClient.on('setLayerVisibility:success', (event) => {
+  console.log('Layer visibility change is complete!');
+});
```
jkissel, line 243: If you return the Promise, there's no need for clients to listen to that event.

jkissel (changes requested):

src/cadenza.js
anistor: Actually, I changed my mind about it , will make it non-optional.

src/cadenza.js
anistor: True that

src/docs.md
anistor: Indeed.

jkissel (approved):

klinakerdisy (approved):

PR #23, opened 2024-03-05
Fix: Conflict when multiple attributes in objectInfo event details ha…

klinakerdisy:
…ve the same printNames.

### DIFF
--- a/src/cadenza.js
+++ b/src/cadenza.js
@@ -1104,7 +1104,7 @@ function array(/** @type unknown */ value) {
 /** @typedef {CadenzaEvent<'editGeometry:ok', {geometry: Geometry}>} CadenzaEditGeometryOkEvent - When the user submitted the geometry. */
 /** @typedef {CadenzaEvent<'editGeometry:cancel'>} CadenzaEditGeometryCancelEvent - When the user cancelled the geometry editing. */
 /** @typedef {CadenzaEvent<'error', {type: string, message?: string}>} CadenzaErrorEvent - An error event that is mapped to a {@link CadenzaError} */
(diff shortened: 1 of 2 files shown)


src/cadenza.js
```diff
@@ -1104,7 +1104,7 @@ function array(/** @type unknown */ value) {
 /** @typedef {CadenzaEvent<'editGeometry:ok', {geometry: Geometry}>} CadenzaEditGeometryOkEvent - When the user submitted the geometry. */
 /** @typedef {CadenzaEvent<'editGeometry:cancel'>} CadenzaEditGeometryCancelEvent - When the user cancelled the geometry editing. */
 /** @typedef {CadenzaEvent<'error', {type: string, message?: string}>} CadenzaErrorEvent - An error event that is mapped to a {@link CadenzaError} */
-/** @typedef {CadenzaEvent<'objectInfo', {layer: WorkbookLayerPath, objectInfos: {selectionIndex: number, formattedValues: Record<string, string>}[]}>} CadenzaObjectInfoEvent - When the user opened the object info flyout. */
+/** @typedef {CadenzaEvent<'objectInfo', {layer: WorkbookLayerPath, objectInfos: {selectionIndex: number, elements: {attributePrintName: Record<string, string>, formattedValues: Record<string, string>}[]}}>} CadenzaObjectInfoEvent - When the user opened the object info flyout. */
```
dalvarezmartinez1, line 1107: how is only a type improvement the fix for this bug? Shouldn't there be some change in code as well?

jkissel: The event is sent from Cadenza, i.e. the related code change is there.

klinakerdisy: Like jkissel said.

The type was indeed wrong..

jkissel: It's still wrong, because it's only _one_ `formattedValue`.

jkissel (changes requested):
objectInfo payload type is still wrong.

jkissel (approved):

PR #28, opened 2024-04-24
More detailed specification of the returned geometry in the CadenzaCh…

klschm:
…angeSelectionEvent and the CadenzaSelectObjectsOkEvent : It's a simplified geometry (and not the full geometry)


klinakerdisy (approved):
👍🏻

PR #30, opened 2024-05-07
Allow communication with parent Cadenza window

MrLynx93:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 - `invalidateCaches` option for `CadenzaClient#reload`
+- Allow communication with parent Cadenza window
```
jkissel, line 9: Proposal:
> (Previously, communication was possible only with an embedded Cadenza iframe.)

src/cadenza.js
```diff
@@ -831,6 +844,10 @@ export class CadenzaClient {
   }
 
   #createUrl(/** @type string */ path, /** @type URLSearchParams */ params) {
+    assert(
+      !this.#isEmbeddedInCadenza,
```
jkissel, line 848: This method is called also from fetchData and downloadData where you don't need an embedded Cadenza, so the name of this getter is confusing.

Proposal: Inline this.#baseUrl != null here.

src/cadenza.js
```diff
@@ -154,26 +154,29 @@ export class CadenzaClient {
    * @param {boolean} [options.debug]
    */
   constructor(baseUrl, { debug = false, iframe, webApplication } = {}) {
```
jkissel, line 156: The baseUrl is now also optional. Please update the docs accordingly.

src/cadenza.js
```diff
@@ -154,26 +154,29 @@ export class CadenzaClient {
...
       assert(
         validExternalLinkKey(webApplication),
         `Invalid webApplication parameter: ${webApplication}`,
       );
     }
 
-    // Remove trailing /
-    if (baseUrl.at(-1) === '/') {
-      baseUrl = baseUrl.substring(0, baseUrl.length - 1);
+    if (baseUrl !== undefined) {
```
jkissel, line 164: I'd rather not be so strict about null vs undefined. People might not be aware of the difference and not use TS.

Proposal: baseUrl != null

src/cadenza.js
```diff
@@ -195,6 +198,22 @@ export class CadenzaClient {
...
+    if (this.#isEmbeddedInCadenza) {
+      assert(
+        !(window.parent === window && window.opener === null),
+        'Cannot find parent Cadenza window',
+      );
+      return window.parent;
+    } else {
+      return /** @type {WindowProxy} */ (this.#requiredIframe.contentWindow);
+    }
+  }
```
jkissel, line 215: * Here you're not interested in the URL, so I'd also not check for it.
* `window.parent` is always defined, but it's the same as `window` if there is no parent window.
* If there's an opener, it should also be used.

Proposal:
```javascript
  get #targetWindow() {
    if (this.#iframe) {
      return /** @type {WindowProxy} */ (this.#requiredIframe.contentWindow);
    }

    const targetWindow = window.opener ?? window.parent;
    assert(targetWindow !== window, 'Cannot find target window');
    return targetWindow;
  }
```

src/docs.md
```diff
@@ -69,6 +69,15 @@ If for some reason you don't like the global `cadenza` field, you can remove it
 const cadenza = window.cadenza.noConflict();
 ```
 
+### Communication with parent Cadenza
```
jkissel, line 72: The window. is confusing, imo. Also I'd expect this whole info in the constructor JSDoc.

Maybe we should mention that fact that a method works with opener and parent in its JSDoc?

jkissel (changes requested):

klschm (changes requested):

src/docs.md
MrLynx93: `window.` is also in all other examples right above it, maybe it looks weird here in the review out-of-context. It makes sense if you look at it rendered. I will also add the info to the constructor JSDoc.

src/cadenza.js
MrLynx93: That would be nice, but it would be incompatible change, we would need to release a new major version. (and we want this in 9.4 as well)

src/cadenza.js
MrLynx93: I added all of the suggestions from comments.

src/cadenza.js
jkissel: We can define overloads for the function and deprecate the one with the baseUrl param:
```
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 68a9392..a834a98 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Added
+- `cadenza()` overload with a single options parameter
 - `invalidateCaches` option for `CadenzaClient#reload`
 
+### Deprecated
+- `cadenza()` overload with the `baseUrl` and `options` parameters
+
 ## 2.7.0 - 2024-04-17
 
 ## 2.6.0 - 2024-04-12
diff --git a/src/cadenza.js b/src/cadenza.js
index 8c46d85..c220f6d 100644
--- a/src/cadenza.js
+++ b/src/cadenza.js
@@ -1,20 +1,44 @@
 /**
- * Creates an instance of the Cadenza JS client.
- *
- * @param {string} baseUrl - The base URL of the Cadenza server
- * @param {object} [options] - Options
- * @param {HTMLIFrameElement | string} [options.iframe] - An iframe for embedding Cadenza or the iframe's ID.
+ * @typedef CadenzaClientOptions
+ * @property {string} [baseUrl] - The base URL of the Cadenza server
+ * @property {HTMLIFrameElement | string} [options.iframe] - An iframe for embedding Cadenza or the iframe's ID.
  *   The iframe is required only for methods that embed Cadenza in an iframe, so e.g. not for {@link CadenzaClient#fetchData}.
  *   If you want to embed Cadenza in multiple iframes, you need to create an instance of the `CadenzaClient` per iframe.
- * @param {ExternalLinkKey} [options.webApplication] - An external link that Cadenza uses to resolve the
+ * @property {ExternalLinkKey} [options.webApplication] - An external link that Cadenza uses to resolve the
  *   [target origin](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#targetorigin) when posting events.
  *   This is required if Cadenza and your application are not running on the same origin.
  *   Please ensure that the user has view privilege for that link!
- * @param {boolean} [options.debug] - Whether to enable debug logging
+ * @property {boolean} [options.debug] - Whether to enable debug logging
+ */
+
+/**
+ * Creates an instance of the Cadenza JS client.
+ *
+ * @overload
+ * @param {CadenzaClientOptions} [options] - Options
+ * @return {CadenzaClient}
+ * @throws For invalid arguments
+ */
+/**
+ * Creates an instance of the Cadenza JS client.
+ *
+ * @deprecated This overload is deprecated since <time>2024-05-08</time> and will eventually be removed.
+ *   Please use the other overload.
+ * @overload
+ * @param {string} baseUrl - The base URL of the Cadenza server
+ * @param {Exclude<CadenzaClientOptions, 'baseUrl'>} [options] - Options
+ * @return {CadenzaClient}
  * @throws For invalid arguments
  */
-export function cadenza(baseUrl, options) {
-  return new CadenzaClient(baseUrl, options);
+/**
+ * @param {string | CadenzaClientOptions} [baseUrlOrOptions]
+ * @param {CadenzaClientOptions} [options]
+ */
+export function cadenza(baseUrlOrOptions, options) {
+  if (typeof baseUrlOrOptions === 'string') {
+    options = { baseUrl: baseUrlOrOptions, ...options };
+  }
+  return new CadenzaClient(options);
 }
 
 /* @ts-ignore */
@@ -147,13 +171,9 @@ export class CadenzaClient {
 
   /**
    * @hidden
-   * @param {string} baseUrl
-   * @param {object} [options]
-   * @param {HTMLIFrameElement | string} [options.iframe]
-   * @param {ExternalLinkKey} [options.webApplication]
-   * @param {boolean} [options.debug]
+   * @param {CadenzaClientOptions} [options]
    */
-  constructor(baseUrl, { debug = false, iframe, webApplication } = {}) {
+  constructor({ baseUrl = '', debug = false, iframe, webApplication } = {}) {
     assert(validUrl(baseUrl), `Invalid baseUrl: ${baseUrl}`);
     if (webApplication) {
       assert(

```

klschm: Ah, very nice, yes, good idea @jkissel :+1: 
Can you make this code change @MrLynx93 ?

klschm (approved):

src/cadenza.js
MrLynx93: I added it

src/cadenza.js
```diff
@@ -831,6 +858,7 @@ export class CadenzaClient {
   }
 
   #createUrl(/** @type string */ path, /** @type URLSearchParams */ params) {
+    assert(this.#baseUrl != null);
```
jkissel, line 861: The assertion message is missing. Proposal:
> Required base URL is not present.

jkissel (changes requested):

jkissel (approved):

PR #36, opened 2024-07-11
CADENZA-36890 Add the object info to cadenza.js and sandbox.html

JensLuebke:
- Add the object info to cadenza.js and sandbox.html
- Change the selection api to accept multi-dimensional array

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 2.11.0 - 2024-07-12
 ### Added
 - Support for `dataType=pdf` in `CadenzaClient#downloadData` and `CadenzaClient#fetchData`
-- `CadenzaClient#fetchObjectInfo`
 
 ### Fixed
 - Make Cadenza return an error instead of showing an error page.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `CadenzaClient#fetchObjectInfo`
 
+### Fixed
+- Make Cadenza return an error instead of showing an error page.
+
+
 ## 2.10.0 - 2024-06-11
 ### Added
 - `CadenzaClient#destroy`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Added
+- Support for `dataType=pdf` in `CadenzaClient#downloadData` and `CadenzaClient#fetchData`
 - `CadenzaClient#fetchObjectInfo`
 
 ### Fixed
 - Make Cadenza return an error instead of showing an error page.
-
 
 ## 2.10.0 - 2024-06-11
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+## 2.11.0 - 2024-07-12
 ### Added
 - Support for `dataType=pdf` in `CadenzaClient#downloadData` and `CadenzaClient#fetchData`
 - `CadenzaClient#fetchObjectInfo`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- `CadenzaClient#fetchObjectInfo`
 
 ## 2.11.0 - 2024-07-12
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- `CadenzaClient#fetchObjectInfo`
 
 ## 2.10.0 - 2024-06-11
 ### Added
--- a/sandbox.html
+++ b/sandbox.html
@@ -509,8 +509,8 @@
     <input name="layers" id="layers">
     <small>
       <p>
-        A JSON value like <code>["layerPrintName",...]</code> or
-        <code>[["layerGroupPrintName","layerPrintName"],...]</code>
+        A JSON value like <code>["&lt;layerPrintName&gt;",...]</code> or
+        <code>[["&lt;layerGroupPrintName&gt;","&lt;layerPrintName&gt;"],...]</code>
       </p>
     </small>
   </div>
@@ -544,7 +544,7 @@
     <label for="objectids">List of List of Object IDs *</label>
     <textarea name="objectIds" id="objectids" rows="5" required></textarea>
     <small>
-      <p>A JSON value like <code>[["object1-id"],["object2-id"]]</code></p>
+      <p>A JSON value like <code>[["&lt;object1-id&gt;"],["&lt;object2-id&gt;"]]</code></p>
     </small>
   </div>
 </template>
@@ -555,8 +555,8 @@
     <input name="layer" id="layer" required>
     <small>
       <p>
-        A JSON value like <code>"layerPrintName"</code> or
-        <code>["layerGroupPrintName","layerPrintName"]</code>
+        A JSON value like <code>"&lt;layerPrintName&gt;"</code> or
+        <code>["&lt;layerGroupPrintName&gt;","&lt;layerPrintName&gt;"]</code>
       </p>
     </small>
   </div>
--- a/sandbox.html
+++ b/sandbox.html
@@ -173,8 +173,8 @@
       },
       fetchObjectInfo: data => {
         console.log('Inspect the fetchObjectInfo() request in the devtools.');
-        cadenzaClient.fetchObjectInfo(data.embeddingTargetId, JSON.parse(data.layer), JSON.parse(data.objectIds), getOptions(data));
-
+        cadenzaClient.fetchObjectInfo(data.embeddingTargetId, JSON.parse(data.layer), JSON.parse(data.objectIds), getOptions(data))
+          .then(oi => console.log(oi));
       },
       downloadData: data => cadenzaClient.downloadData(data.embeddingTargetId, data.dataType, getOptions(data))
     };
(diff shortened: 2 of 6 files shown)


jkissel:
If you need these changes in Cadenza 9.4, please create another PR for the 2.8 branch, which is used there.

(So for 9.4 you should create a new version 2.8.2 and for 9.5, 9.x, 10.0 and 10x a new version 2.12.0.)

sandbox.html
```diff
@@ -531,12 +539,12 @@
 </template>
 <template data-action="closeMe">No Parameters needed.</template>
 
-<template id="common-layer-selection">
+<template id="common-object-ids">
   <div>
-    <label for="values">List of Object IDs *</label>
-    <textarea name="values" id="values" rows="5" required></textarea>
+    <label for="objectids">List of List of Object IDs *</label>
```
jkissel, line 544: Proposal to avoid ambiquity:
> List of object ID value lists

sandbox.html
```diff
@@ -572,6 +580,25 @@
...
+  <div>
+    <label>
+      <input type="checkbox" name="useMapSrs">
+      Use the map SRS instead of WGS84
+    </label>
+  </div>
+  <div>
+    <label>
+      <input type="checkbox" name="fullGeometries">
+      Return the non-simplified geometry
```
jkissel, line 597: geometry -> geometries

sandbox.html
```diff
@@ -572,6 +580,25 @@
...
 
+<template data-action="fetchObjectInfo" data-common="layer,object-ids,filter">
+  <div>
+    <label for="embeddingTargetId">Embedding target ID *</label>
+    <input name="embeddingTargetId" id="embeddingTargetId" required>
+  </div>
+  <div>
+    <label>
+      <input type="checkbox" name="useMapSrs">
+      Use the map SRS instead of WGS84
```
jkissel, line 591: I think we should describe this option consistently throughout the sandbox.

src/cadenza.js
```diff
@@ -844,19 +844,54 @@ export class CadenzaClient {
...
+   *   Currently only map views are supported.
+   * @param {(WorkbookLayerPath | string)[]} [layerPath] - Layer path to identify the layer
+   *  (identified using layer paths or print names)
+   * @param {unknown[][]} objectIds - The IDs of the objects to select
+   * @param {object} [options] - Options
+   * @param {FilterVariables} [options.filter] - Filter variables
+   * @param {AbortSignal} [options.signal] - A signal to abort the data fetching
+   * @param {Boolean} [options.useMapSrs] - Use the map SRS instead of WGS84
+   * @param {Boolean} [options.fullGeometries] - Return non-simplified geometries
+   * @return {Promise<Response>} A `Promise` for the fetch response
```
jkissel, line 860: It's required to return the plain Response instead of a FeatureCollection?

src/cadenza.js
```diff
@@ -844,19 +844,54 @@ export class CadenzaClient {
...
+   * @throws For invalid arguments
+   */
+  fetchObjectInfo(source, layerPath, objectIds, { filter, signal, useMapSrs, fullGeometries } = {}) {
+    this.#log('CadenzaClient#fetchObjectInfo', ...arguments);
+    const params = createParams({
+      filter});
+    return this.#fetch(
+      resolvePath(source) + "/objectinfo" ,
```
jkissel, line 868: Please use the development watch (npm start) when developing Cadenza JS. The CI pipeline will not pass with code formatting issues.

src/cadenza.js
```diff
@@ -844,19 +844,54 @@ export class CadenzaClient {
...
+      JSON.stringify({ objectIds, layerPath: array(layerPath), useMapSrs, fullGeometries}),
+      "application/json");
+  }
+
   async #fetch(
     /** @type string */ path,
     /** @type URLSearchParams */ params,
     /** @type AbortSignal | undefined */ signal,
+    /** @type String | undefined If body is set, the fetch will be a post.*/ body,
+    /** @type String | undefined Has to be set if body is set.*/ bodyContentType,
```
jkissel, line 880: I think for now we can assume the Content-Type to be JSON if a body is given.

sandbox.html
```diff
@@ -531,12 +539,12 @@
...
-<template id="common-layer-selection">
+<template id="common-object-ids">
   <div>
-    <label for="values">List of Object IDs *</label>
-    <textarea name="values" id="values" rows="5" required></textarea>
+    <label for="objectids">List of List of Object IDs *</label>
+    <textarea name="objectIds" id="objectids" rows="5" required></textarea>
     <small>
-      <p>A JSON value like <code>["object1","object2"]</code></p>
+      <p>A JSON value like <code>[["object1-id"],["object2-id"]]</code></p>
```
jkissel, line 547: But it worked with the simple array ...?! Since this API is not published, this is not a breaking change, but why making it less convenient for the default case in maps where we have only one ID value?

src/cadenza.js
```diff
@@ -1058,6 +1093,7 @@ function assertSupportedDataType(
  * @param {TablePart[]} [params.parts]
  * @param {'MAP'} [params.targetType]
  * @param {boolean} [params.useMapSrs]
+ * @param {unknown[][]} [params.objectIds]
```
jkissel, line 1096: Unnecessary change, right? The IDs are never passed via params.

src/docs.md
```diff
@@ -352,6 +352,21 @@ const tableData = await response.json();
 ...
 ```
 
+### Download the Object Info from a Workbook Map View
```
jkissel, line 355: Download -> Fetch

src/cadenza.js
```diff
@@ -844,19 +844,54 @@ export class CadenzaClient {
     return this.#fetch(resolvePath(source), params, signal);
   }
 
+  /**
+   * Fetch object info from a workbook view.
+   *
+   * @param {EmbeddingTargetId} source - The workbook view to fetch object info from.
+   *   Currently only map views are supported.
```
jkissel, line 851: Proposal: Change the main description above instead of mentioning this limitation only here. We do not expect the object info to be supported for other views, right?

> Fetch the object info from a workbook map view.

jkissel (changes requested):

sandbox.html
JensLuebke: What is the non map srs option, if set to false? Here it is WGS84

jkissel: It's _always_ WGS84, that's why I'm confusing that it's mentioned here explicitly. Imo mentioning the default SRS in the api docs is enough.

JensLuebke: okay. thanks for putting that clear. I will change it

sandbox.html
JensLuebke: Before the single array was send to the backend and one value was applied on all ids of the array of ids.... so the multi id selection was broken anyway

src/cadenza.js
```diff
@@ -133,6 +133,18 @@ globalThis.cadenza = Object.assign(
  * _Note:_ Since numbers in JavaScript are Double values ([more info on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding)),
  * for Long variables, the API is currently limited to the Double value range.
  */
+/**
+ * @typedef Feature - A [GeoJSON](https://geojson.org/) feature object
+ * @property {any[]} id - The id of the feature
```
jkissel, line 138: This definition is not according to the [spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2):
> If a Feature has a commonly used identifier, that identifier SHOULD be included as a member of the Feature object with the name "id", and the value of this member is either a JSON string or number.

Maybe we should model this as a [foreign member](https://datatracker.ietf.org/doc/html/rfc7946#page-15) instead, e.g. "cadenzaObjectId"?

src/cadenza.js
```diff
@@ -133,6 +133,18 @@ globalThis.cadenza = Object.assign(
...
+ * @property {Geometry} geometry - The geometry
+ * @property {Record<string, string>} properties - The properties
+ * @property {Number | undefined} area - The area of the polygon
+ * @property {Number | undefined} length - The length of the line
```
jkissel, line 142: I think this should be like this:
```javascript
 * @property {number} [area] - The area of a `Polygon` feature
 * @property {number} [length] - The area of a `LineString` feature
```
* Primitive types are lower case.
* Optional props are marked with [].

src/cadenza.js
```diff
@@ -133,6 +133,18 @@ globalThis.cadenza = Object.assign(
...
+ */
+/**
+ * @typedef FeatureCollection - A [GeoJSON](https://geojson.org/) feature collection object
+ * @property {Feature[]} features - The type of the geometry
```
jkissel, line 146: Wrong description

src/cadenza.js
```diff
@@ -133,6 +133,18 @@ globalThis.cadenza = Object.assign(
  * _Note:_ Since numbers in JavaScript are Double values ([more info on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding)),
  * for Long variables, the API is currently limited to the Double value range.
  */
+/**
+ * @typedef Feature - A [GeoJSON](https://geojson.org/) feature object
+ * @property {any[]} id - The id of the feature
+ * @property {Geometry} geometry - The geometry
+ * @property {Record<string, string>} properties - The properties
```
jkissel, line 140: Maybe mention that it's the formatted object info?

src/cadenza.js
```diff
@@ -868,24 +879,23 @@ export class CadenzaClient {
...
   ) {
     const url = this.#createUrl(path, params);
     this.#log('Fetch', url.toString());
     const method = body ? "POST" : undefined;
     const headers = new Headers();
     headers.set('X-Requested-With', 'XMLHttpRequest');
-    if (bodyContentType) {
-      headers.set('Content-Type', bodyContentType);
+    if (body) {
+      headers.set('Content-Type', "application/json");
```
jkissel, line 898: You still do not use the development watch, right? The CI will not allow to merge this.

src/docs.md
```diff
@@ -364,7 +364,7 @@ const response = await cadenzaClient.fetchObjectInfo('embeddingTargetId', 'layer
   fullGeometries: true
 });
 
-const objectInfo = await response.json();
+const objectInfo = await response;
```
jkissel, line 367: The request already _is_ the objectInfo now.

jkissel (changes requested):

sandbox.html
jkissel: :+1:

src/cadenza.js
JensLuebke: I can add it as a foreign member. Then i would not provide a actually ID, since i do not have one.
Right?

CHANGELOG.md
```diff
@@ -8,10 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `CadenzaClient#fetchObjectInfo`
 
+## 2.11.0 - 2024-07-12
+### Added
+- Support for `dataType=pdf` in `CadenzaClient#downloadData` and `CadenzaClient#fetchData`
+- `CadenzaClient#fetchObjectInfo`
```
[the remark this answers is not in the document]
JensLuebke: merging error. sorry

src/cadenza.js
jkissel: Yes, I think that would be better to not conflict with the spec :+1:

JensLuebke: Andreas have a opinion about the id. Perhaps we discuss this again on Monday

JensLuebke: we choose the name objectId

jkissel (approved):

PR #37, opened 2024-08-27
CADENZA-36891: Extend API for geometry intersection

AndreasWBartels:


sandbox.html
```diff
@@ -611,6 +618,42 @@
...
+  </div>
+  <div>
+    <label>
+      <input type="checkbox" name="useMapSrs">
+      Use map SRS
+    </label>
+  </div>
+  <div>
+    <label for="distance">Buffer size</label>
+    <input name="distance" id="distance">
```
jkissel, line 643: Proposal: Add type="number" and min="0" here.

sandbox.html
```diff
@@ -611,6 +618,42 @@
...
+    <label for="lengthUnit">Length unit</label>
+    <select name="lengthUnit" id="lengthUnit">
+      <option>m</option>
+      <option>km</option>
+    </select>
+    <small>
+      <p>
+        A positive buffer size is needed for point or line intersection geometries.
```
jkissel, line 651: Should we add an assertion for the distance to be positive?

src/cadenza.js
```diff
@@ -108,6 +108,15 @@ globalThis.cadenza = Object.assign(
...
  * */
 
+/**
+ * @typedef Distance
+ * @property {number} value
+ * @property {LengthUnit} lengthUnit
+ */
+/**
+ * @typedef {'m'|'km'} LengthUnit
+ */
```
jkissel, line 118: Should this be "Buffer"?
```javascript
/**
 * @typedef Buffer
 * @property {number} value
 * @property {'m'|'km'} unit
 */
```
Should we make the unit optional (assuming "m" as the default)?

src/cadenza.js
```diff
@@ -919,6 +928,42 @@ export class CadenzaClient {
...
 
+  /**
+   *  Fetch the intersection area from a workbook map view layer in JSON format for a given area.
+   *
+   * @param {EmbeddingTargetId} source - The workbook view to fetch object info from.
+   * @param {(WorkbookLayerPath | string)[]} layerPath - Layer path to identify the layer
+   *  (identified using layer paths or print names)
+   * @param {Geometry} geometry - The intersection geometry
+   * @param {object} [options] - Options
+   * @param {boolean} [options.useMapSrs]  -  Whether the geometry and the extent are in the map's SRS (otherwise EPSG:4326 is assumed)
```
jkissel, line 939: This method does not have an "extent" option.

src/cadenza.js
```diff
@@ -919,6 +928,42 @@ export class CadenzaClient {
...
+   * @param {Distance} [options.buffer] - Buffer size for geometry of the transition
+   * @param {AbortSignal} [options.signal] - A signal to abort the data fetching
+   * @return {Promise<FeatureCollection>} A `Promise` for the fetch response
+   * @throws For invalid arguments
```
jkissel, line 943: Atm, we do not do any argument validations.

src/cadenza.js
```diff
@@ -919,6 +928,42 @@ export class CadenzaClient {
...
+   * @server
+   */
+  fetchAreaIntersection(
+    source,
+    layerPath,
+    geometry,
+    { useMapSrs, buffer, signal} = {},
+  ) {
+    this.#log('CadenzaClient#areaIntersection', ...arguments);
+    const params = createParams({});
```
jkissel, line 953: Proposal: Rather make the "params" argument optional instead of passing a dummy value.
```javascript
@type URLSearchParams | undefined
```

src/docs.md
```diff
@@ -367,6 +367,30 @@ const response = await cadenzaClient.fetchObjectInfo('embeddingTargetId', 'layer
 const objectInfo = await response;
 ```
 
+### Fetch the area intersection from a Workbook Map View Layer
+
+<small>API: [CadenzaClient#fetchObjectInfo(./classes/CadenzaClient.html#fetchAreaIntersection)</small>
+
+Download the intersection area from a workbook map view layer in JSON format for a given area. The result contains all intersecting objects of the layer, their object ID, if defined the object name and a geometry representing the intersection area, with area size and area perimeter.
```
jkissel, line 374: In Cadenza JS "Download" means that a download dialog is triggered in the browser. So this is "Fetch".

src/docs.md
```diff
@@ -367,6 +367,30 @@ const response = await cadenzaClient.fetchObjectInfo('embeddingTargetId', 'layer
...
+const geometry = {
+  type: 'Point',
+  coordinates: [328_627.563458, 5_921_296.662223],
+};
+const bufferSize = {
+  value: 100,
+  lengthUnit: 'm'
+}
+
+const response = await cadenzaClient.fetchAreaIntersection('embeddingTargetId', 'layerPrintName', geometry, {
```
jkissel, line 386: Most of the examples in here use &lt;placeholders>. I'd keep it consistent (and also fix the OI example).

src/docs.md
```diff
@@ -367,6 +367,30 @@ const response = await cadenzaClient.fetchObjectInfo('embeddingTargetId', 'layer
...
+  useMapSrs: true,
+  bufferSize: bufferSize
+});
+
+const featureCollection = await response;
```
jkissel, line 391: Doesn't make sense to await the response twice (also in the OI example).

jkissel (changes requested):

sandbox.html
AndreasWBartels: min 0 is only needed if the intersection geometry is a point or linestring

sandbox.html
AndreasWBartels: polygons can have a negative distance

sandbox.html
```diff
@@ -293,7 +301,6 @@
           <optgroup label="Without Iframe">
             <option value="downloadData">Download data</option>
             <option value="fetchData">Fetch data</option>
-            <option value="fetchObjectInfo">Fetch Object Info</option>
```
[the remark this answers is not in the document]
AndreasWBartels: merge mistake

sandbox.html
```diff
@@ -611,6 +618,42 @@
...
+  <div>
+    <label for="embeddingTargetId">Embedding target ID *</label>
+    <input name="embeddingTargetId" id="embeddingTargetId" required>
+  </div>
+  <div>
+    <label for="geometry">Intersection Geometry (GeoJSON) *</label>
+    <textarea name="geometry" id="geometry" rows="5" required></textarea>
+    <small>
+      <p>
+        The geometry must result in an area in combination with the buffer specification.
```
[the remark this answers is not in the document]
AndreasWBartels: If we want a assertion a geometry type validation would be necessary here. 
Now you get an error response, if there is no buffer in the request.

sandbox.html
jkissel: For polygons we also support negative distances!?

sandbox.html
jkissel: Yeah, but why not? We have that info in the method.

src/cadenza.js
AndreasWBartels: optional is "* @property {'m'|'km'} [unit] ", but how to set a default value on the property?

jkissel: This should work:
```
 * @property {'m' | 'km'} [unit='m']
```

sandbox.cjs
```diff
@@ -6,8 +6,8 @@ const DEFAULT_CADENZA_URL = 'http://localhost:8080/cadenza';
 
 const index = process.argv.indexOf(CADENZA_URL_PARAM);
 const cadenzaUrlFromArgs = index !== -1
-    ? process.argv[index + 1]
-    : process.argv.find(arg => arg.startsWith(CADENZA_URL_PARAM + '='))?.slice(CADENZA_URL_PARAM.length + 1);
+  ? process.argv[index + 1]
+  : process.argv.find(arg => arg.startsWith(CADENZA_URL_PARAM + '=')).slice(CADENZA_URL_PARAM.length + 1);
```
jkissel, line 10: The ? in ?.slice( is still missing.

sandbox.html
```diff
@@ -244,7 +251,8 @@
         layers: layers ? JSON.parse(layers) : undefined,
         useMapSrs: useMapSrs === 'on',
         fullGeometries: fullGeometries === 'on',
-        zoomTarget: zoomToGeometry === 'on' ? { type: 'geometry' } : undefined
+        zoomTarget: zoomToGeometry === 'on' ? { type: 'geometry' } : undefined,
+        buffer: distance ? { value: Number(distance), lengthUnit: lengthUnit ? lengthUnit : 'm' } : undefined
```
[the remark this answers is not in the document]
jkissel: I meant that
```javascript
lengthUnit: lengthUnit ? lengthUnit : 'm'
```
can be simplified to
```javascript
lengthUnit
```
because the `lengthUnit` cannot be undefined.

sandbox.html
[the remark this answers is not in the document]
jkissel: So far we try to assert things if possible _before_ hitting the server.

src/docs.md
jkissel: Why not "just" assign the result to that variable straight-away?
```javascript
const featureCollection = await ...
```

jkissel (changes requested):

sandbox.html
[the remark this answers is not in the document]
AndreasWBartels: in cadenza.js?

jkissel (approved):

JensLuebke (commented):
approved

PR #43, opened 2024-09-24
CADENZA-37227 feat: Add 'action' event, which is fired as 'action' event when clicking a postMessage action Cadenza link in views shown via `CadenzaClient#show`

sebastianbarth-disy:
https://jira.disy.net/browse/CADENZA-37227

This adds a 'action' event analogously to 'drillthrough'. 'drillthrough' postmessages are sent always with a `values` object and an optional `context` string. The new `action` event only has a `context` string sent along with the pm.

'action' postmessages are sent when clicking an 'postMessage action' link. Those links are defined per External URL in the Management Center and can be used in Text, Indicator and Image views in Workbooks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `CadenzaClient#fetchAreaIntersections`
 - `CadenzaClient#fetchObjectInfo`
+- `CadenzaActionEvent`
 
 ## 2.11.0 - 2024-07-12
 ### Added
--- a/src/cadenza.js
+++ b/src/cadenza.js
@@ -291,7 +291,9 @@ export class CadenzaClient {
    * @param {AbortSignal} [options.signal] - A signal to abort the iframe loading
    * @return {Promise<void>} A `Promise` for when the iframe is loaded
    * @throws For invalid arguments
-   * @fires {@link CadenzaDrillThroughEvent}
+   * @fires
+   * - {@link CadenzaDrillThroughEvent}
+   * - {@link CadenzaActionEvent}
    * @embed
    */
   show(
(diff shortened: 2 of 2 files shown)


jkissel:
> The context to be optional (Designed as such to make it similar to 'drillthrough' - It is also optional on the UI).
What's the point of an action if you don't know which one it is?

> Despite being organized per URL, the URL is absolutely irrelevant here (same situation with existing postMessage Drillthroughs).
It's not completely irrelevant, the URL is used to derive the target origin.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `CadenzaClient#fetchAreaIntersections`
 - `CadenzaClient#fetchObjectInfo`
+- `CadenzaActionEvent`, which is fired as 'action' event when clicking a postMessage action Cadenza link in views shown via `CadenzaClient#show`
```
jkissel, line 11: Usually we leave the details to the apidocs to not have it in two places. For example the event is fired also for #showMap, but it's also not mentioned here.

src/cadenza.js
```diff
@@ -291,7 +291,7 @@ export class CadenzaClient {
    * @param {AbortSignal} [options.signal] - A signal to abort the iframe loading
    * @return {Promise<void>} A `Promise` for when the iframe is loaded
    * @throws For invalid arguments
-   * @fires {@link CadenzaDrillThroughEvent}
+   * @fires {@link CadenzaDrillThroughEvent | CadenzaActionEvent}
```
jkissel, line 294: Does this show up as individual links to the docs of these classes in the docs? Otherwise please do it like e.g. with the #editGeometry events.

src/cadenza.js
```diff
@@ -1311,6 +1313,9 @@ function array(/** @type unknown */ value) {
  * @property {TYPE} type - The event type
  * @property {DETAIL} detail - Optional event details (depending on the event type)
  */
+/**
+ * @typedef {CadenzaEvent<'action', {context?: string}>} CadenzaActionEvent - When the user executed a POST message action.
```
jkissel, line 1317: Proposal: "... which is defined on an external link in the Cadenza management center."

jkissel (changes requested):

PR #47, opened 2024-10-08
CADENZA-36955 CADENZA-37861 feat: Added `setCustomValidity()` and `Va…

sebastianbarth-disy:
The customer is enabled to implement additional custom validation of geometries in reaction of a `geometry:update` postMessage event sent when the user changes the geometry in a geometry editor dialog (started via `createGeometry()` or `editGeometry()` calls).

In this situation the customer can use the new `setCustomValidity(message, type)` function to add an additional error, warn, info, success state:
```
  /**
   * Set custom validity state of the geometry editor in addition to the default validation state (including errors and
   * warnings). When set to error the dialog submission is blocked.
   * If there already is a custom state set it will override it.
   * Passing '' will reset the custom state to undefined, meaning no custom state is displayed.
   * If no geometry editing is started, the method call has no effect.
   * @param message {string} The message to show in the dialog
   * @param [type] {CustomValidityType} The type of message (defaults to 'error')
   */
  setCustomValidity(message, type = 'error') {
    ...
  }
  
  /** @typedef {'error'|'warning'|'info'|'success'} CustomValidityType - The type of custom validity used for disclosing custom validity state */
```

image

### DIFF
--- a/sandbox.html
+++ b/sandbox.html
@@ -181,7 +181,8 @@
         cadenzaClient.fetchAreaIntersections(data.embeddingTargetId, JSON.parse(data.layer), JSON.parse(data.geometry), getOptions(data))
           .then(console.log);
       },
-      downloadData: data => cadenzaClient.downloadData(data.embeddingTargetId, data.dataType, getOptions(data))
+      downloadData: data => cadenzaClient.downloadData(data.embeddingTargetId, data.dataType, getOptions(data)),
+      setCustomValidity: data => cadenzaClient.setCustomValidity(data.message, data.type || undefined)
     };
 
     const form = document.getElementById('form');
@@ -297,6 +298,7 @@
             <option value="createGeometry">Create Geometry</option>
             <option value="editGeometry">Edit Geometry</option>
             <option value="selectObjects">Select Objects</option>
+            <option value="setCustomValidity">Set Custom Validity</option>
           </optgroup>
           <optgroup label="Without Iframe">
             <option value="downloadData">Download data</option>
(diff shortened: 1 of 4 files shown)


sebastianbarth-disy:
Pull request is meant for both, 10.1 and 10.2-planned. Since current master already contains changes meant for 10.2-planned only, we first need to branch off a version for 10.1 and release the 10.2-planned-only changes with a higher minor version (e.g. 2.14.0).

sandbox.html
```diff
@@ -143,7 +143,7 @@
       iframe: parentMode ? undefined : 'iframe',
       debug: true
     });
-    const actionHandlers = {
+    window.actionHandlers = {
```
jkissel, line 146: Why global?

sandbox.html
```diff
@@ -472,9 +473,28 @@
...
 
+<template id="common-setCustomValidity" >
+  <fieldset>
+    <legend>setCustomValidity</legend>
+    <label for="customValidityType">Type</label>
+    <select name="customValidityType" id="customValidityType">
+      <option>success</option>
+      <option>info</option>
+      <option>warning</option>
+      <option selected="selected">error</option>
```
jkissel, line 484: Note: The value is not needed, because "selected" is a boolean attribute.

sandbox.html
```diff
@@ -472,9 +473,28 @@
...
+      type: document.getElementById('customValidityType').value,
+      message: document.getElementById('customValidityMessage').value
+    });">Send</button>
+  </fieldset>
+</template>
+
 <template data-action="setFilter" data-common="filter"></template>
 
-<template data-action="createGeometry" data-common="map,filter">
+<template data-action="createGeometry" data-common="map,filter,setCustomValidity">
```
jkissel, line 497: It's a separate method, so it should also have it's own action in the actions dropdown instead of being included in the the create/editGeometry actions.

src/cadenza.js
```diff
@@ -648,6 +649,23 @@ export class CadenzaClient {
...
   }
 
+  /**
+   * Set custom validity state of the geometry editor in addition to the default validation state (including errors and
+   * warnings). When set to error the dialog submission is blocked.
+   * If there already is a custom state set it will override it.
+   * Passing '' will reset the custom state to undefined, meaning no custom state is displayed.
+   * If no geometry editing is started, the method call has no effect.
+   * @param message {string} The message to show in the dialog
+   * @param [type] {CustomValidityType} The type of message (defaults to 'error')
```
jkissel, line 659: The {type} comes first.

jkissel (changes requested):

sandbox.html
tedbarth: Used in line 188 in a button inline.

sandbox.html
jkissel: Yes, correct :+1:

jkissel (approved):

PR #60, opened 2024-10-29
Slb/cadenza 37480

ivigilm:


sandbox.html
```diff
@@ -476,7 +476,7 @@
   </div>
   <div>
     <label for="geometry">Geometry (GeoJSON)</label>
-    <textarea name="geometry" id="geometry" rows="5"></textarea>
+    <textarea name="geometry" id="geometry" rows="5" placeholder="{&quot;coordinates&quot;:[[[9.598504509838506,52.70992820638244],[8.899110702389294,51.72257425885121],[10.906499617840836,51.725944168909365],[9.598504509838506,52.70992820638244]]],&quot;type&quot;:&quot;Polygon&quot;}"></textarea>
```
jkissel, line 479: Placeholders are not very user-friendly, because they disappear when focusing the input. In other places (e.g. "Filter") we use a small hint below the input instead.

src/cadenza.js
```diff
@@ -623,7 +623,7 @@ export class CadenzaClient {
    * @param {OperationMode} [options.operationMode] - The mode in which a workbook should be operated
    * @param {ZoomTarget} [options.zoomTarget] - A target Cadenza should zoom to
    * @param {AbortSignal} [options.signal] - A signal to abort the iframe loading
-   * @param {Object[]} [options.additionalLayers] - Layer definitions to be imported and shown in the background, as a basis for the drawing. Each is a layer definition, with name, type and content (a Geojson featureCollection).
+   * @param {LayerDefinition[]} [options.additionalLayers] - Layer definitions to be imported and shown in the background, as a basis for the drawing. Each is a layer definition, with name, type and content (a Geojson featureCollection).
```
jkissel, line 626: :+1:

jkissel (approved):

sandbox.html
ivigilm: Do we have any preferences here, then, for the sandbox? Because in some fields we do have placeholders still.
In this case I used them because for me it looked nicer, and I was thinking users would only need an initial hint. It also looked good enough for Klaus, who had requested the improvement. Since you approved the merge I'd keep it like this for now but will keep it in mind, if that's ok.

jkissel: In Cadenza, UX just discouraged the use of placeholders completely. But yeah, fine for me.

PR #61, opened 2024-11-05
Slb/cadenza 37667

qulol:


CHANGELOG.md
```diff
@@ -5,7 +5,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project uses a version scheme based on the Cadenza main version in the format x.x.y, where x.x is the Cadenza main version and y a functional change or bugfix.
 
 ## Unreleased
-
+### Changed
+- `CadenzaEditGeometryUpdateEvent`. 
+  - Updated the type definition for the `editGeometry:update` event.
+    - The event's `detail` payload has been changed from `{ geometry: Geometry }` to `FeatureCollection | Feature | undefined`, to bring them in line with the rest of our API.
```
jkissel, line 11: Needs resolution / answers:
* Why undefined instead of an empty collection?

Proposals:
* This is a breaking change. I'd suggest to mark them specifically, now that we do not have semantic versioning anymore, e.g. using a **BREAKING CHANGE** prefix on the bullet point.
* "Changed" and "Updated" and ChangeEditGeometryUpdate and editGeometry:update are redundant. Proposal:
> The `CadenzaEditGeometryUpdateEvent#detail` is now typed as `FeatureCollection | Feature | undefined` to align it with the rest of the API. (Before: `Geometry`)
* Please check that this indentation actually ends up in the [Releases](https://github.com/DisyInformationssysteme/cadenza.js/releases/tag/v10.2.2) (which is what customers look at).

src/cadenza.js
```diff
@@ -118,9 +118,11 @@ globalThis.cadenza = Object.assign(
  * @typedef {'m'|'km'} LengthUnit
  */
 
+/** @typedef {number[]} Coordinate - A tuple with an x and y value // [number, number] */
```
jkissel, line 121: The type should be `[number, number]` instead of `number[]`. Then you also don't need to mention that in the description.

src/cadenza.js
```diff
@@ -153,14 +155,16 @@ globalThis.cadenza = Object.assign(
  */
 /**
  * @typedef Feature - A adapted [GeoJSON](https://geojson.org/) feature object.
+ * @property {'Feature'} type - The object's type
  * @property {any[]} objectId - The id of the feature
  * @property {Geometry} geometry - The geometry
  * @property {Record<string, string>} properties - The formated properties
  * @property {number} [area] - The area of a `Polygon` feature
- * @property {number} [length] - The area of a `LineString` feature
+ * @property {number} [length] - The length of a `LineString` feature
```
jkissel, line 163: :+1:

src/cadenza.js
```diff
@@ -153,14 +155,16 @@ globalThis.cadenza = Object.assign(
  */
 /**
  * @typedef Feature - A adapted [GeoJSON](https://geojson.org/) feature object.
```
jkissel, line 157: Typo: A**n**

CHANGELOG.md
```diff
@@ -5,7 +5,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- `Feature`
+  - Updated the type definition.
+    - Added a `type` property, so it corresponds better to the actual used type.
+    - Added an optional `circumference` property, so it corresponds better to the actual used type.
+- `FeatureCollection`
+  - Updated the type definition.
+    - Added a `type` property, so it corresponds better to the actual used type.
+- `Geometry`
+  - Updated the type definition.
+    - Added a `coordinates` property, so it corresponds better to the actual used type.
```
jkissel, line 24: All theses "Added" things should go in the "Added" section:

```
### Added
- `Feature#type` and `#circumference`
- `FeatureCollection#type`
- `Geometry#coordinates`
```
I think that's also enough for the changelog, because you can read the API docs to find out whether it's optional.

jkissel (changes requested):

CHANGELOG.md
qulol: > Why undefined instead of an empty collection?
Since FeatureCollection is a special usecase and will not occur in case of multi-geometries, it does not make sense to use them if the multi-geometry feature is absent.

jkissel (approved):

PR #62, opened 2024-11-05
 TENT-5976 Introduce *-dev versions in code and establish a "Branch off Major Version" script

sebastianbarth-disy:
See: https://im.disy.net/channel/dev-cadenzajs?msg=X8hjBfWKmu9wSDNRg

* Use the version number of the non-released version in all branches/only main branch
* Use -dev suffix for unreleased git versions
* Split-off branch creation for major releases

### DIFF
--- a/.github/workflows/branchOffCadenzaVersion.yml
+++ b/.github/workflows/branchOffCadenzaVersion.yml
@@ -1,0 +1,66 @@
+name: Branch off Cadenza Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      cadenza-version:
+        type: string
+        required: true
+        description: |
+          Next Cadenza Main Version
+          (e.g. '10.2' for setting the main branch up for development against Cadenza 10.2. This would make 10.2.0-dev the new main branch version.)
+
+
+jobs:
+  release:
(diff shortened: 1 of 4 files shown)


.github/workflows/branchOffCadenzaVersion.yml
```diff
@@ -0,0 +1,66 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      cadenza-version:
+        type: string
+        required: true
+        description: |
+          Next Cadenza Main Version
+          (e.g. '10.2' for setting `main` branch version to 10.2.0-dev as corresponding major-minor version of in Cadenza 10.2.)
```
jkissel, line 11: "of in" does not look right. Proposal:
> e.g. '10.2' for setting the main branch up for development against Cadenza 10.2. This would make 10.2.0-dev the new main branch version.

jkissel (approved):

.github/workflows/branchOffCadenzaVersion.yml
sebastianbarth-disy: Thanks, will adopt it!

PR #66, opened 2024-11-08
TENT-5976 fix: Remove v prefix for NPM RELEASE_TAG

sebastianbarth-disy:
This bug lead to NPM use a tag name "vlatest" instead of "latest".

It might be enough to remote this NPM tag on npmjs.com and update "latest" TAG to use 10.2.3 instead of 10.2.2.


jkissel (approved):
I fixed the tags in the npm registry.

PR #71, opened 2024-11-25
CADENZA-37531 and CADENZA-38242:

klschm:
- Handle 'layerData' extentStrategy correctly
- Reformat 'locationFinder' example in sandbox


src/cadenza.js
```diff
@@ -1469,6 +1469,8 @@ function sanitizeExtentStrategy({
           return { type: 'geometry', geometry };
         }
         break;
+      case 'layerData':
```
jkissel, line 1472: Maybe just
```
default:
```
?

jkissel (approved):

src/cadenza.js
klschm: Oh, right :+1:

PR #74, opened 2024-11-26
Slb/ad/cadenza 38410

qulol:


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project uses a version scheme based on the Cadenza main version in the format x.x.y, where x.x is the Cadenza main version and y a functional change or bugfix.
 
 ## Unreleased
+### Fixed
+- Removed duplicate geometry section in `sandbox.html#editGeometry`
+- Horizontal scrollbar in `sandbox.html` left sidebar now appears sticky.
```
jkissel, line 10: A Horizontal scrollbar is rather painful to use. Why do we want it here?

jkissel (changes requested):

CHANGELOG.md
qulol: We currently have it due to long explaination texts, and this is a improvement to the current state imo.

jkissel: But the sidebar is fixed width. We should be able to craft the contents in a way that does not require a horizontal scrollbar. For example texts should wrap lines. We can have a look if necessary.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project uses a version scheme based on the Cadenza main version in the format x.x.y, where x.x is the Cadenza main version and y a functional change or bugfix.
 
 ## Unreleased
+### Fixed
+- Removed duplicate geometry section in `sandbox.html#editGeometry`
+
+### Changed
+- Examples are now displayed more compact
```
klschm, line 12: - Examples in the sandbox now are displayed more compact

klschm (approved):

CHANGELOG.md
jkissel: Or:
> Improved the examples in the sandbox to avoid a horizontal scrollbar.

jkissel (approved):

PR #76, opened 2025-01-02
RPINTERN-849: Feat: Implement zooming in the setSelection method

samuelSandru:


sandbox.html
```diff
@@ -161,7 +161,7 @@
       },
       setFilter: data => cadenzaClient.setFilter(parseFilterVariables(data.filter)),
       setLayerVisibility: data => cadenzaClient.setLayerVisibility(JSON.parse(data.layer), data.visibility === 'on'),
-      setSelection: data => cadenzaClient.setSelection(JSON.parse(data.layer), JSON.parse(data.objectIds)),
+      setSelection: data => cadenzaClient.setSelection(data.layer, {values: JSON.parse(data.objectIds), extentStrategy: JSON.parse(data.extentStrategy)}),
```
klschm, line 164: Imo the 1st parameter like in the other methods should be
`JSON.parse(data.layer)`
instead of the currently done
`data.layer`

src/cadenza.js
```diff
@@ -499,13 +505,42 @@ export class CadenzaClient {
    *
    * @hidden
    * @param {WorkbookLayerPath | string} layer - The data view layer to set the selection in
-   * @param {unknown[][]} values - The IDs of the objects to select
+   * @param {{values: unknown[][], extentStrategy?: GeometryExtentStrategy} | unknown[][]} [options] - Options or values to select.
+   * @param options.values - The IDs of the objects to select
+   * @param [options.extentStrategy] - The extentStrategy to zoom to
```
klschm, line 510: The `options` object needs to be named as `__namedParameters` when it is deconstructed in the method (like it was already done in main in multiple methods), because otherwise the parameters are not described well in the generated documentation.

src/cadenza.js
```diff
@@ -499,13 +505,42 @@ export class CadenzaClient {
    *
    * @hidden
    * @param {WorkbookLayerPath | string} layer - The data view layer to set the selection in
-   * @param {unknown[][]} values - The IDs of the objects to select
+   * @param {{values: unknown[][], extentStrategy?: GeometryExtentStrategy} | unknown[][]} [options] - Options or values to select.
```
klschm, line 508: Here it's said with the rectangle brackets around the `options` (meaning `[options]`), that the options are optionally available. But below it's asserted, that the `values` parameter inside the `options` needs to be available. So the `options` are not optionally but mandatory. Please remove the rectangle brackets around the `options` to also provide that information in the method signature

src/cadenza.js
```diff
@@ -499,13 +505,42 @@ export class CadenzaClient {
...
+
+  /**
+   * The extent to zoom to.
+   *
+   * @param {GeometryExtentStrategy} [extentStrategy]
+   */
+  #setExtentStrategy(extentStrategy) {
+    const type = extentStrategy?.type;
+    // Other extent strategies are handled via URL parameters.
+    if (type === 'geometry' || type === 'layerData') {
```
klschm, line 541: The type `layerData` doesn't exist in this scenario, so the check for it can be removed

klschm (changes requested):

src/cadenza.js
jkissel, line 510: Proposal:
> Strategy to define the new map extent

(You cannot zoom to a strategy.)

src/cadenza.js
```diff
@@ -499,13 +505,42 @@ export class CadenzaClient {
...
+      'Please make sure to provide at least one ID for the object you want to select',
+    );
+    const { values, extentStrategy } = options;
     this.#log('CadenzaClient#setSelection', ...arguments);
-    return this.#postRequest('setSelection', { layer: array(layer), values });
+    const setSelectionResult = await this.#postRequest('setSelection', {
+      layer: array(layer),
+      values,
+    });
+    this.#setExtentStrategy(extentStrategy);
```
jkissel, line 529: Why not include the extendStrategy in the "setSelection" event? (Saves one roundtrip.)

jkissel (changes requested):

jkissel:
* Please add a CHANGELOG, e.g.
  ```markdown
  ### Added
  * `extentStrategy` option for the internal methods `CadenzaClient#setSelection`, `CadenzaClient#addSelection`, and `CadenzaClient#removeSelection`
  * Please also merge this change to the 10.2 branch.

src/cadenza.js
```diff
@@ -136,6 +136,12 @@ globalThis.cadenza = Object.assign(
  * @property {'geometry'} type The type of the zoom target
  */
 
+/**
+ * @typedef GeometryExtentStrategy - The given {@link Geometry} defines the initial map extent.
```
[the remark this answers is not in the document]
samuelSandru: I wanted to be as closed as possible to the 10.2 code this is why I did it this way

sandbox.html
```diff
@@ -161,7 +161,7 @@
...
       addSelection: data => cadenzaClient.addSelection(JSON.parse(data.layer), JSON.parse(data.objectIds)),
       removeSelection: data => cadenzaClient.removeSelection(JSON.parse(data.layer), JSON.parse(data.objectIds)),
```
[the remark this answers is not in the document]
samuelSandru: I'll gladly do it if there's some extra budget provided. Btw removeSelection does not need this or?

src/cadenza.js
```diff
@@ -499,13 +505,42 @@ export class CadenzaClient {
...
    * @return {Promise<void>} A `Promise` for when the selection was set.
    * @postMessage
    */
-  setSelection(layer, values) {
+  async setSelection(layer, options = { values: [] }) {
+    // For backwards compatibility, the options can be simply the values
+    if (Array.isArray(options)) {
+      options = { values: options };
+    }
+    assert(
```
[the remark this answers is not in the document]
samuelSandru: I hope I don't need to touch the other methods, there's no requirement and no budget: If budget and requirements appear I will do that for the other methods as well

sandbox.html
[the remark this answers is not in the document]
jkissel: Did you talk to your PO?

src/cadenza.js
samuelSandru: because this is how it was done by @klschm on 10.2 and I wanted to be spend as little budget as possible and be as similar to that solution as possible to make cherry-picking trivial

src/cadenza.js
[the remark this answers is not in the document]
jkissel: Even if we do not add the new option to the other methods, we should keep our API consistent. So either remove the assertion or also add it to the other methods. And the assertion should be done after the logging.

src/cadenza.js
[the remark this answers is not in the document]
jkissel: Still it doesn't make sense to have two types for the same thing.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- `extentStrategy` option for the internal methods `CadenzaClient#setSelection`, `CadenzaClient#addSelection`, `CadenzaClient#removeSelection`
+
+### Changed
+- **BREAKING CHANGE** Renamed `zoomTarget` parameter, `ZoomTarget` type and `GeometryZoomTarget` type to `extentStrategy`, `ExtentStrategy` and `GeometryExtentStrategy`
```
klschm, line 13: @jkissel We now do the breaking change (regarding "zoomTarget" -> "extentStrategy") also for 10.1:
- Should we then instead of doing a "patch" version release (which would result in 2.14.2, because 2.14.1 was the release before) do a "minor" version release (which would result in 2.15.0)? If we would do so, should we then also create the v2.15.x branch and mention that one (instead of v2.14.x) in our internal confluence page describing which Cadenza JS branch fits to which Cadenza branch?
- Or should we use a "patch" version release, even if we do a breaking change?

src/cadenza.js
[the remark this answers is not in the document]
klschm: @jkissel We decided to do the breaking change also for 10.1 - See the actual code state. I hope, this is fine?

klschm (approved):

src/cadenza.js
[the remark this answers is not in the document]
jkissel: I don't really feel comfortable a breaking change, which is not actually needed. But according to the custom app responsibles, zoomTarget is still unused, so the breaking change is hopefully OK.

Since in pre-10.2 branches we use semantic versioning, this would be v3.0.0 then.

CHANGELOG.md
jkissel: I don't really feel comfortable a breaking change, which is not actually needed. But according to the custom app responsibles, zoomTarget is still unused, so the breaking change is hopefully OK.

Since in pre-10.2 branches we use semantic versioning, this would be v3.0.0 then. From my side, we can also create the branch when we need it in the future, but please update the wiki page.

jkissel (approved):

CHANGELOG.md
klschm: @samuelSandru So please...
- create a new branch "v3.0.0" after merging (see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-and-deleting-branches-within-your-repository )
- then release this new branch for the first time (3.0.0 will be created)
- update and commit&push the Maven_Parent/pom.xml of Cadenza 10.1
- update the Cadenza JS confluence so that it is documented, that Cadenza 10.1 now uses 3.x
-

samuelSandru: yeah let's make sure i don't fuck up anything :D

PR #84, opened 2025-04-04
CADENZA-39576: Introduced UiFeature to disable dialog footers

AndreasWBartels:


AndreasWBartels:
Provisional commit, a button-specific activation may be necessary as a result of the conversion to series production.

sandbox.html
```diff
@@ -377,6 +377,7 @@
         <li>workbook-design</li>
         <li>workbook-map-add-layer</li>
         <li>workbook-view-management</li>
+        <li>workbook-view-dialog-footer-buttons</li>
```
jkissel, line 380: I think this is not particularly related to workbook views and I still think (as described in the estimation) that it would be good to prefix embedding-only features with embedding-. Proposal: embedding-buttons.

jkissel (changes requested):

sandbox.html
jkissel: I still don't see how this is related to views. What if we add another dialog to the embedding API, which is not view-related - Would you add another feature?

sandbox.html
```diff
@@ -400,6 +400,9 @@
         <li>workbook-design</li>
         <li>workbook-map-add-layer</li>
         <li>workbook-view-management</li>
+        <li>embedded-dialog-footer-buttons</li>
```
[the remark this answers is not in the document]
AndreasWBartels: footer because we disable the buttons placed in the footer and no who are placed inside the dialog.

jkissel (approved):

PR #85, opened 2025-04-30
Add support for disabling `geometry-edit-translate` UIFeature

qulol:


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project uses a version scheme based on the Cadenza main version in the format x.x.y, where x.x is the Cadenza main version and y a functional change or bugfix.
 
 ## Unreleased
+### Added
+- UIFeature `geometry-edit-translate` has been introduced to allow disabling translation interactions when editing geometries.
```
klschm, line 9: Imo we should not mention implementation details of Cadenza in the changelog of Cadenza JS because they can change and this UIFeature maybe is also valid for other places in Cadenza in the future.
Proposal:
- Support for 'geometry-edit-translate' as disabledUiFeatures value

klschm (changes requested):

CHANGELOG.md
jkissel: Actually, we have this type also in Cadenza JS: https://disyinformationssysteme.github.io/cadenza.js/types/UiFeature.html But the new feature is not described there.

When this is done, we do not need the description here and this would be enough:
```
### Added
- `UiFeature` value `"geometry-edit-translate"`
```

jkissel: Actually, we have this type also in Cadenza JS: https://disyinformationssysteme.github.io/cadenza.js/types/UiFeature.html But the new feature is not described there.

When this is done, we do not need the description here and this would be enough:
```
### Added
- `UiFeature` value `"geometry-edit-translate"`
```

jkissel (changes requested):

CHANGELOG.md
klschm: :+1: You are completely right - This is the much better approach!

klschm (approved):

jkissel (approved):

PR #92, opened 2025-05-19
Slb/cadenza 40215

klschm:


src/cadenza.js
```diff
@@ -1296,7 +1296,7 @@ function validKebabCaseString(/** @type string */ value) {
 }
 
 function validRepositoryName(/** @type string */ value) {
-  return /^[\w -]{1,255}$/.test(value);
+  return /^[\wäöüÄÖÜß -]{1,255}$/.test(value);
```
jkissel, line 1299: Why do we want to allow invalid names here? This is the technical repository name, not the print name.

PR #97, opened 2025-05-20
CADENZA-40236 feat: Add support for setting spatial filter on embedding targets

sebastianbarth-disy:


src/cadenza.js
```diff
@@ -183,13 +183,24 @@ globalThis.cadenza = Object.assign(
  */
 /** @typedef {'columns' | 'values' | 'totals'} TablePart - A part of a table to export */
 /**
- * @typedef {Record<string, string | string[] | number | Date | null>} FilterVariables - Filter variable names and values
+ * @typedef {Record<string, string | string[] | number | Date | SpatialFilterValue | null>} FilterVariables - Filter variable names and values
  *
  * Variables of type String, Integer, Long, Double and Date can be set.
+ * A property named "$spatial" reflects a Spatial filter value and must adhere to type {@link SpatialFilterValue}.
```
sebastianbarth-disy, line 189: We cannot combine a Record<> with a normal interface, which is why we have to allow SpatialFilterValue for all via type and validate at runtime that it only happens for $spatial property.

sandbox.html
```diff
@@ -390,6 +390,33 @@
       <p>A JSON object with filter variable names and values</p>
       <p>Dates need to be ISO strings<br>(e.g. <code>"2023-11-17T17:12:06.175Z"</code>).<br>Use <code>"{{currentDate}}"</code> for the current date.</p>
       <p>Pass multiple values for an <code>IN</code> expression as <code>["a", "b", ...]</code></p>
+      <p>Spatial filter values can be set using the dedicated <code>"$spatial"</code> variable name:
+        <ul>
+          <li>The value is an object aligning to the following structure:<br>
```
jkissel, line 395: Proposal: "aligning to" -> "with"

sandbox.html
```diff
@@ -390,6 +390,33 @@
...
+          [8.89, 51.72],
+          [10.90, 51.72],
+          [9.59, 52.70]
+        ]
+      ],
+      "type": "Polygon"
+    }
+  }
+}</pre></li>
+          <li>The <code>"spatialRelation"</code> property accepts the values:<ul><li><code>"overlaps"</code></li><li><code>"notOverlaps"</code></li>
```
jkissel, line 413: Indentation?

src/cadenza.js
```diff
@@ -183,13 +183,24 @@ globalThis.cadenza = Object.assign(
...
  * for Long variables, the API is currently limited to the Double value range.
  */
+/**
+ * @typedef SpatialFilterValue - Defines a spatial filter variable value.
+ * @property {SpatialRelation} spatialRelation - Defines how geometries must interact with the given filter geometry to be accepted by the filter
+ * @property {SpatialFilterGeometry} geometry - A valid [GeoJSON](https://geojson.org/) geometry restricted to type "Polygon".
+ */
+/** @typedef {'overlaps' | 'notOverlaps' | 'contains'} SpatialRelation - Defines how geometries interact */
+/**
+ * @typedef {Geometry} SpatialFilterGeometry
```
jkissel, line 201: Maybe rather define a Polygon type etc? More reusable ...

jkissel (approved):

sandbox.html
sebastianbarth-disy: Improved.

PR #101, opened 2025-06-23
CADENZA-40037 feature: Add 'layout' URL parameter to specify UI layout

jpalys:


sandbox.html
```diff
@@ -107,7 +107,7 @@
   </style>
 
   <script type="module">
-    import { cadenza } from './cadenza.js';
+    import { cadenza } from './src/cadenza.js';
```
jkissel, line 110: Mistake?

jkissel (changes requested):

sandbox.html
jpalys: no, actually this should be like that to work locally

jkissel: But now it does not work anymore in production. To use this without a sandbox running in Cadenza, you need to run `npm run sandbox`.

jkissel: Yes, this should stay as it is, but you don't need to modify it for development. Instead, you should start the development watch in parallel (npm start). This puts the tsc output in "." where it's picked up by the sandbox.

jkissel (approved):

PR #107, opened 2025-07-04
CADENZA-40405 Removed deprecated options locationFinder and mapExtent

qulol:


src/cadenza.js
```diff
@@ -1687,36 +1647,27 @@ function array(/** @type unknown */ value) {
...
-        const extent = extentStrategy.extent ?? mapExtent;
+        const extent = extentStrategy.extent;
         if (extent) {
           return { type: 'static', extent };
         }
         break;
       case 'locationFinder':
-        const query = extentStrategy.query ?? locationFinder;
+        const query = extentStrategy.query;
         if (query) {
```
qulol, line 1671: We don't marked is as optional parameter, can we remove the if ?

src/cadenza.js
```diff
@@ -181,8 +181,6 @@ globalThis.cadenza = Object.assign(
  * @property {UiFeature[]} [disabledUiFeatures] - Cadenza UI features to disable
```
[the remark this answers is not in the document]
klschm: It's described in the "geometry" parameter description of line 160:  When you define one, it `...takes precedence over another geometry that might be given in an API call.`. If you don't define one, the one from the API call is used

src/cadenza.js
klschm: We cannot assume, the customers use TS - So they may use JS only and call the extentStrategy without a query or extent filled and that should be covered by our code

src/cadenza.js
```diff
@@ -1687,36 +1647,27 @@ function array(/** @type unknown */ value) {
...
  * @param {ExtentStrategy} [__namedParameters.extentStrategy]
  * @return {ExtentStrategy | undefined}
  */
-function sanitizeExtentStrategy({
-  extentStrategy,
-  mapExtent,
-  locationFinder,
-  geometry,
-} = {}) {
+function sanitizeExtentStrategy({ extentStrategy, geometry } = {}) {
```
klschm, line 1660: Did you configure your IDE correctly? This should not be formatted like it is now, but like it was done before for better readability imo...

src/cadenza.js
```diff
@@ -971,22 +943,10 @@ export class CadenzaClient {
...
-      extentStrategy,
-      filter,
-      layers,
-      locationFinder,
-      mapExtent,
-      useMapSrs,
-      operationMode,
-      signal,
-    } = {},
+    { extentStrategy, filter, layers, useMapSrs, operationMode, signal } = {},
```
klschm, line 946: Did you configure your IDE correctly? This should not be formatted like it is now, but like it was done before for better readability imo...

src/cadenza.js
qulol: But we have jsdoc where it is mentioned what we expect? It doesn't make sense to allow having a ExtentStrategy of type LocationFinder and having the query param not specified.

src/cadenza.js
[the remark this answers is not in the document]
qulol: Same argument here: If an ExtentStrategy is explicitly set with type geometry, It should be required that the geometry is set aswell...
If you don't want to use a custom ExtentStrategy, then leave this extentStrategy out entirely!?

src/cadenza.js
qulol: It is less then 80 characters, so the formatting is correct imo

src/cadenza.js
qulol: Yes, we should definitely validate this on the client and throw an error if no query param is set, since then the LocationFinderExtentStrategy is simply not configured correctly.
Same argument for the GeometryExtentStrategy which even has the geometry parameter declared as optional...

jkissel (approved):

PR #108, opened 2025-07-10
CADENZA-40418 update version to 10.4.2-dev because of a broken releas…

andkoons:
…e build which resulted in version 10.4.1 already to be used in the npm registry

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@disy/cadenza.js",
-  "version": "10.4.1-dev",
+  "version": "10.4.2-dev",
   "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


jkissel (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@disy/cadenza.js",
-  "version": "10.4.1-dev",
+  "version": "10.4.2-dev",
```
klschm, line 3: This has also to be done in the package-lock.json etc.
Please run `npm install` so see which files change too

klschm (changes requested):

klschm (approved):

PR #110, opened 2025-07-23
CadenzaEditGeometryOkEvent typing and examples

qulol:


klschm (changes requested):

src/docs.md
```diff
@@ -285,6 +285,53 @@ cadenzaClient.on('editGeometry:cancel', (event) => {
 });
 ```
 
+### Edit multiple Geometries
+
+<small>API: [CadenzaClient#batchEditGeometry](./classes/CadenzaClient.html#batchEditGeometry), [CadenzaClient#on](./classes/CadenzaClient.html#on)</small>
+
+Edit a collection of geometries with a workbook map view in the background. The geometry coordinates are in the map's SRS (`useMapSrs: true`).
+The editor will create an additional layer where editable geometries are stored and selectable for editing. Additional geometries can also be created from within the editor once it has been initialized.
```
jkissel, line 293: Isn't this line more like implementation details? Or is this somehow relevant (or even seeable) for the user?

jkissel (approved):

src/docs.md
qulol: The layer is visible for the user and he can toggle visibility and select features from it. It acts like a regular layer on the map. For my taste, it makes sense to describe it.

klschm: My opinion on this, like already discussed some weeks ago: It is visible, yes, but it is not part of the API itself but an implementation detail, how we reach the goal to let the user create and/or edit multiple features with the same API call

klschm (approved):

